PR #17, opened 2021-10-09
chore: add publishConfig and registry

targeral:


packages/cli/babel-preset-app/package.json
```diff
@@ -51,5 +51,8 @@
     "output": {
       "packageMode": "node-js"
     }
+  },
+  "publishConfig": {
+    "registry": "https://registry.npmjs.org/"
```
caohuilin, line 56: 还需要 access public

PR #77, opened 2021-10-15
fix(server): fix dev options in production env

zllkjc:


packages/server/server/src/server/dev-server.ts
```diff
@@ -23,6 +24,19 @@ import Watcher from '../dev-tools/watcher';
...
 
+const DEFAULT_DEV_OPTIONS: DevServerOptions = {
+  client: {
+    port: '8080',
+    overlay: false,
+    logging: 'none',
+    path: HMR_SOCK_PATH,
+    host: 'localhost',
+  },
+  https: false,
```
await-ovo, line 35: https ?

PR #273, opened 2021-11-05
fix: fix language detect in Windows

rqzheng2015:
For some unknown reasons, process.env.LC_ALL || process.env.LC_MESSAGES || process.env.LANG || process.env.LANGUAGE gets 'undefined' at Windows Node enviorment, so we add  Intl.DateTimeFormat().resolvedOptions().locale methods for detecting user locale.

reference:
1. https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/DateTimeFormat/resolvedOptions
2. https://stackoverflow.com/questions/46072248/node-js-how-to-detect-user-language

### DIFF
--- a/.changeset/six-experts-stare.md
+++ b/.changeset/six-experts-stare.md
@@ -1,0 +1,5 @@
+---
+"@modern-js/i18n-cli-language-detector": patch
+---
+
+fix locale detect in Windows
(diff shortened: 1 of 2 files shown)


caohuilin (approved):
LGTM.

PR #297, opened 2021-11-11
feat: reduce lint time at CI workflow

rqzheng2015:
As now eslint process will cover all files in projects, and it took too much time, we aim to reduce lint time by only lint changed files compared to origin/main branch

### DIFF
--- a/package.json
+++ b/package.json
@@ -19,6 +19,7 @@
     "test": "pnpm run test --filter ./packages",
     "prepare": "pnpm run prepare --filter ./packages",
     "lint": "modern lint",
+    "lint:check": "modern lint --no-fix",
     "change": "modern change",
     "clear": "modern clear",
     "deploy": "modern deploy",
--- a/packages/cli/plugin-jarvis/src/lint.ts
+++ b/packages/cli/plugin-jarvis/src/lint.ts
@@ -35,7 +35,13 @@ export default () => {
     rawArgs.push(...ensureOption(args, 'ignore-pattern', pattern));
   });
 
-  if (args?._?.length) {
+  // To speed up eslint, develop can set LINT_CHANGED_FILES=true
+  // to activate eslint only works on changed files compared to origin/main branch
+  if (process.env.LINT_CHANGED_FILES) {
+    rawArgs.push(
+      `--no-error-on-unmatched-pattern $(git diff --name-only origin/main...HEAD | xargs)`,
+    );
+  } else if (args?._?.length) {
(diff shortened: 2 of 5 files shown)


package.json
```diff
@@ -19,6 +19,7 @@
     "test": "pnpm run test --filter ./packages",
     "prepare": "pnpm run prepare --filter ./packages",
     "lint": "modern lint",
+    "lint:check": "modern lint --no-fix",
```
xuchaobei, line 22: 直接在 CI 中，通过scripts 中的 lint 命令 + 参数的方式可以实现同样效果吗？ scripts 里一般不应该轻易新增命令

packages/cli/plugin-jarvis/src/lint.ts
```diff
@@ -35,7 +35,13 @@ export default () => {
     rawArgs.push(...ensureOption(args, 'ignore-pattern', pattern));
   });
 
-  if (args?._?.length) {
+  // To speed up eslint, develop can set LINT_CHANGED_FILES=true
+  // to activate eslint only works on changed files compared to origin/main branch
+  if (process.env.LINT_CHANGED_FILES) {
+    rawArgs.push(
+      `--no-error-on-unmatched-pattern $(git diff --name-only origin/main...HEAD | xargs)`,
```
anc95, line 42: 我感觉现阶段在 CI ymal 文件上写

yarn lint -- --no-error-on-unmatched-pattern $(git diff --name-only origin/main...HEAD | xargs) 

是不是更合适一些。

1. LINT_CHANGED_FILES 算是个 API 了，需要讨论下，比如不用环境变量，可以提供一个 yarn lint --git-changed-diff=origin/main
2. main 写死了应该也不合适，这些参数应该都是从 CI 脚本上控制的

rqzheng2015: Thanks for comment. Here's what I'm thinking:
1. "origin/main" like you say, should not be hard code, developers should pass branch name as param.
2. At default situration, eslint will use './' to fully lint all files, so we have to use an enviorment variable like LINT_CHANGED_FILES to determine whether we should add specify args to lint changed files only.  But the name can be still renamed if there are better suggestions.

package.json
rqzheng2015: As command "modern lint" will silencely fix files that are not following the eslint rules, developers can't know what happened exactly unless they check git commits. We should add an extra
check command for testing the files and report the error on console output. So I suggest we add "lint:check"

PR #315, opened 2021-11-17
chore(new-action): fix typo

lianghx-319:


packages/generator/new-action/src/module.ts
```diff
@@ -113,11 +113,11 @@ export const ModuleNewAction = async (options: IModuleNewActionOption) => {
     generator = `${generator}@${distTag}`;
   }
 
-  const devDependencie =
+  const devDependencies =
```
caohuilin, line 116: I think devDependency maybe more suitable?

lianghx-319: Yep.

caohuilin (approved):
LGTM.

PR #336, opened 2021-11-24
Fix webpack base config type error

targeral:


packages/cli/webpack/src/config/base.ts
```diff
@@ -610,7 +610,8 @@ class BaseWebpackConfig {
       ] as any)
       .end()
       .minimizer('css')
-      .use(CssMinimizerPlugin, [
+      // 11.24 发现类型突然出现的问题，临时增加 use<any>解决。
+      .use<any>(CssMinimizerPlugin, [
```
await-ovo, line 614: update @types/css-minimizer-webpack-plugin ?

targeral: 你那边有尝试过嘛？

await-ovo (approved):

PR #363, opened 2021-12-03
feat: support build --analyze

xuchaobei:


packages/cli/webpack/src/config/client.ts
```diff
@@ -280,6 +281,16 @@ class ClientWebpackConfig extends BaseWebpackConfig {
         },
       ]);
     }
+
+    if (this.options.cliOptions?.analyze) {
```
await-ovo, line 285: It may only be better in the production environment

xuchaobei: --analyze  is only related to build command,  not the execution environment.

caohuilin (approved):

PR #397, opened 2021-12-17
fix(sass): remove @types/sass

rqzheng2015:
Since "sass" has already provided type definition since 1.45.0, we remove @types/sass dependencies to avoid incorrect type warnings.


packages/toolkit/esmpack/package.json
```diff
@@ -57,7 +57,7 @@
     "resolve": "~1.19.0",
     "rollup": "~2.50.6",
     "rollup-plugin-node-polyfills": "~0.2.1",
-    "sass": "~1.32.2",
+    "sass": "~1.45.0",
```
targeral, line 60: 这个最好还是恢复成原来的样子。它这里通过 ~ 限制了版本再 1.32.2 - 1.(32+1).0

rqzheng2015: LGTM

targeral (approved):

PR #422, opened 2021-12-28
fix: app init function not work

caohuilin:


packages/runtime/src/compatible.tsx
```diff
@@ -60,8 +60,8 @@ export const createApp = ({ plugins }: CreateAppOptions) => {
               runner.init(
                 { context: contextValue },
                 {
-                  onLast: ({ context: context1 }) => {
-                    (App as any)?.init?.(context1);
+                  onLast: async ({ context: context1 }) => {
+                    await (App as any)?.init?.(context1);
```
tqma113, line 64: 这里只需要 `return (App as any)?.init?.(context1)` 就可以了

packages/runtime/src/compatible.tsx
```diff
@@ -109,8 +109,8 @@ export const bootstrap: BootStrap = async (
     runner.init(
       { context: _context },
       {
-        onLast: ({ context: context1 }) => {
-          (App as any)?.init?.(context1);
+        onLast: async ({ context: context1 }) => {
+          await (App as any)?.init?.(context1);
```
tqma113, line 113: 这里也是，只需要 return

xuchaobei (approved):

PR #475, opened 2022-01-13
feat: add *.global.less/sass/scss support

rqzheng2015:
**Goal:** 
add *.global.less/sass/scss support when output.disableCssModuleExtension is set to true.
**Info:** 
When we follow the [tourrial](https://modernjs.dev/docs/guides/usages/css/css-modules#%E5%85%A8%E9%9D%A2%E5%90%AF%E7%94%A8-css-modules) and set `output.disableCssModuleExtension` to true, all css/less/sass files will be considered as css modules files if they were not named as `[name].global.css`.
```css
/* app.global.css */
.bg-blue {
  background-color: blue;
}
```
But css support is **not enough**, the developers might also want to use less/sass/scss for further uses.
So this PR provides the ability to fulfill their wishes:)

### DIFF
--- a/.changeset/flat-goats-argue.md
+++ b/.changeset/flat-goats-argue.md
@@ -1,0 +1,7 @@
+---
+"@modern-js/plugin-less": patch
+"@modern-js/plugin-sass": patch
+"@modern-js/webpack": patch
+---
+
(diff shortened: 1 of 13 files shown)


rqzheng2015:
> add test case

Okay, boss.

targeral (dismissed):

caohuilin (approved):

PR #488, opened 2022-01-15
fix(lint): add file filter ACM [skip ci]

zoolsher:
Eslint got an error for file not existing when a file is deleted in develop branch.


zoolsher:
@leeight PTAL

leeight:
Good catch.

PR #520, opened 2022-01-21
fix: fix esbuild error when project's tsconfig.json sets "target:es5"

rqzheng2015:
# PR Details

fix esbuild error when project's tsconfig.json sets "target:es5"

## Description

Problem: When use modern.config.js for configuration in modern.js project, which has some typings that ES5 doesn't support, like `const`
```typescript
// modern.config.ts
const a=1;
```
and tsconfig.json has been set as “target:es5"
```typescript
{
  "compilerOptions": {
    "target": "es5",
  }
}
```

The Compiler will throw error like this:
```shell
Transforming const to the configured target environment ("es5") is not supported yet
```

Reason:
According to ESbuild 0.12.6 [change note](https://github.com/evanw/esbuild/releases/tag/v0.12.6), `target` would be read from 'tsconfig.json' when esbuild's own target is not defined

Fix:
Manually set esbuild's target to `esnext` in @modern-js/node-bundle-require

Some snippets for the feature:
Only respect target in tsconfig.json when esbuild's target is not configured (#1332)
In version 0.12.4, esbuild began respecting the target setting in tsconfig.json. However, sometimes tsconfig.json contains target values that should not be used. With this release, esbuild will now only use the target value in tsconfig.json as the language level when esbuild's target setting is not configured. If esbuild's target setting is configured then the target value in tsconfig.json is now ignored.

## Types of changes

## Checklist

- [x] All new and existing tests passed.

### DIFF
--- a/packages/toolkit/node-bundle-require/src/index.ts
+++ b/packages/toolkit/node-bundle-require/src/index.ts
@@ -66,6 +66,10 @@ export async function bundleRequire(filepath: string, options?: Options) {
     format: 'cjs',
     platform: 'node',
     bundle: true,
+    // fix transforming error when the project's tsconfig.json
+    // sets `target: "es5"`
+    // reference: https://github.com/evanw/esbuild/releases/tag/v0.12.6
+    target: 'esnext',
     ...options?.esbuildOptions,
     plugins: [
       ...(options?.esbuildPlugins || []),
--- a/.changeset/lucky-birds-laugh.md
+++ b/.changeset/lucky-birds-laugh.md
@@ -1,0 +1,5 @@
+---
+"@modern-js/node-bundle-require": patch
+---
+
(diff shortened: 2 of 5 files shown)


packages/toolkit/node-bundle-require/src/index.ts
```diff
@@ -66,6 +66,10 @@ export async function bundleRequire(filepath: string, options?: Options) {
     format: 'cjs',
     platform: 'node',
     bundle: true,
+    // fix transforming error when the project's tsconfig.json
+    // sets `target: "es5"`
+    // reference: https://github.com/evanw/esbuild/releases/tag/v0.12.6
+    target: 'esnext',
```
leeight, line 72: add test case

rqzheng2015: done.

xuchaobei (approved):

PR #532, opened 2022-01-25
fix: export all interface from @modern-js/core/config

chenhuang444:
maintain 'export type {...}' to core/config/index.ts
this way core/index.ts can just export via: export * from './config'

# PR Details

## Types of changes

## Checklist

- [x] I have read the **CONTRIBUTING** document.
- [x] All new and existing tests passed.

### DIFF
--- a/packages/cli/core/src/index.ts
+++ b/packages/cli/core/src/index.ts
@@ -20,13 +20,7 @@ import { enable } from '@modern-js/plugin/node';
 
 import type { Hooks } from '@modern-js/types';
 import { program, Command } from './utils/commander';
-import {
-  resolveConfig,
-  defineConfig,
-  loadUserConfig,
-  UserConfig,
-  ToolsConfig,
-} from './config';
+import { resolveConfig, loadUserConfig } from './config';
 import { loadPlugins } from './loadPlugins';
 import {
   AppContext,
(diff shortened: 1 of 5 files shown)


packages/cli/core/src/index.ts
```diff
@@ -26,6 +26,7 @@ import {
   loadUserConfig,
   UserConfig,
   ToolsConfig,
+  DeployConfig,
```
rqzheng2015, line 29: If DeployConfig is added, ,maybe you could also export other Config at one time.
```javascript
export * from './config';
```

chenhuang444: the original export utilizes Typescript `export type { ... }` to ensure all related code is removed at runtime.
see [Type Only Imports and Export](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-8.html#type-only-imports-and-export) for more details.

I've moved the maintenance of 'export type' to core/config/index.ts. This will allow us to use `export * from './config'` while making sure all types are removed at runtime. This will also be better for maintenance since contributors will be able to see the `export` usage in the same file and won't need to modify other files

rqzheng2015: Nice work out there:)

rqzheng2015 (approved):

PR #1357, opened 2022-07-12
doc: avoid prepare failure

LittleMoonkk:
when running pnpm install, always occur prepare failure which cause application can not work

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -19,7 +19,8 @@ npm install -g pnpm@6
 ### Set up local Modern.js repository
 
 ```zsh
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -20,6 +20,7 @@ npm install -g pnpm@6
 
 ```zsh
 pnpm install
```
chenjiahan, line 22: pnpm install 可以改成 pnpm install --ignore-scripts，这样就不会自动执行 prepare 了

chenjiahan (approved):

PR #1498, opened 2022-08-12
feat(web-builder): `createStubBuilder` cache parameters of each hooks

Asuka109:
# PR Details

## Description

In before, the second one will never resolve because memoized build function was already settled.
if we disable the cache of build function, the second one may re-run build in some cases.

```ts
await builder.unwrapHook('onBeforeBuildHook');
await builder.unwrapHook('onBeforeBuildHook');
```

By now `unwrapHook` will always return cached parameters of hooks,
cleaning the cache by run `reset` manually.

## Types of changes

- [x] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist


packages/builder/web-builder/src/shared/utils.ts
```diff
@@ -76,3 +76,14 @@ export function getDataUrlCondition(dataUriLimit = DEFAULT_DATA_URL_SIZE) {
     return source.length <= dataUriLimit;
   };
 }
+
+export function removeErrorStackHead(stack: string, count: number) {
+  const [heading, ...body] = stack.split('\n');
+  return [heading, ...body.slice(count)].join('\n');
+}
+
+export function catchErrorStack(message: string, skip: number) {
```
chenjiahan, line 85: 这个函数在哪用的

Asuka109: `logger::debug` 用到的混到这里来了，我给删掉

Asuka109: 这个感觉还是得用第三方依赖来做，手动处理太麻烦了

chenjiahan (approved):

PR #1863, opened 2022-10-12
feat(builder): improve progress bar

chenjiahan:
# PR Details

## Description

- Refactor the Progress plugin to be a `Class` instead of `Function`, so we can use webpack-chain to register it.
- Fix the progress finished status is never reached.
- Adjust the progress colors priority.
- Removed the condition of `total < 0` because total should not be minus in current implementation.
- Remove the legacy webpackbar plugin because current progress bar is good enough.

Preview:

截屏2022-10-12 16 25 20

## Types of changes

- [x] New feature (non-breaking change which adds functionality)

## Checklist


packages/builder/builder-webpack-provider/src/webpackPlugins/ProgressPlugin/helpers/bar.ts
```diff
@@ -76,14 +72,7 @@ export const renderBar = (option: Partial<Props>) => {
...
   )(cliTruncate(message, messageWidth, { position: 'start' }));
-  const frameStr = fc(frames[current % frames.length]);
-  if (total <= 0) {
-    // 无法获取百分比时用动画表示进度
-    if (columns >= MIDDLE_WIDTH) {
-      return [idStr, frameStr, space, msgStr].join('');
-    }
-    return [idStr, frameStr].join('');
-  }
+
```
chenjiahan, line 75: I think `total` will not be minus yet, so the `total <= 0` condition is removed.

zhusjfaker (approved):

PR #1978, opened 2022-10-28
fix(builder): missing source map of inlined chunks

chenjiahan:
# PR Details

## Description

Fix missing source map of inlined chunks, there are warnings in chrome devtools when some scripts are inlined:

截屏2022-10-28 11 44 38

Changes:

1. Use`delete compilation.assets` instead of `compilation.deleteAsset`, because we want to preserve the related files such as source map.
2. If we inlined the chunk to HTML,we should update the value of sourceMappingURL comment, because the relative path of source code has been changed.

## Types of changes

## Checklist


jkzing (approved):
LGTM 👍

PR #2065, opened 2022-11-09
fix(builder): should emit app icon to static/image dir

chenjiahan:
# PR Details

## Description

If the `html.appIcon` config is used, we should emit app icon to `dist/static/image/icon.png` rather than `dist/icon.png`.

## Types of changes

## Checklist


packages/builder/builder-webpack-provider/src/webpackPlugins/HtmlAppIconPlugin.ts
```diff
@@ -1,31 +1,35 @@
 import fs from 'fs';
-import path from 'path';
+import path, { join } from 'path';
```
fireairforce, line 2:
```suggestion
import {  dirname, join } from 'path';
```

fireairforce (approved):

packages/builder/builder-webpack-provider/src/webpackPlugins/HtmlAppIconPlugin.ts
chenjiahan: done

PR #2168, opened 2022-11-22
feat: fallback logic for streaming SSR

xuchaobei:
# PR Details

## Description

Fallback logic for streaming SSR: 
1. If error occurs before `onShellReady` being called, the whole app would fallback to csr.
2. If error occurs after `onShellReady` being called, only related components would fallback to csr. React would handle this situation automaticlly, so we just need to record the error log. 

## Types of changes

- [x] New feature (non-breaking change which adds functionality)

## Checklist


packages/runtime/plugin-runtime/src/ssr/serverRender/helmet.ts
```diff
@@ -44,7 +45,6 @@ export default function helmet(content: string, helmetData: HelmetData) {
     ${noscript}
     ${script}
     ${style}
-    ${existTitle ? '' : title}
```
zllkjc, line 47: 这句删掉的话，如果 html 中原本不存在 title，helmet 的 title 就加不上了吧

xuchaobei: 是的

zllkjc (approved):

PR #2288, opened 2022-12-05
feat: support internal metaName_TARGET env vars

zllkjc:
# PR Details

In some cases, the code needs to distinguish between SSR and CSR in order to remove environment-related code.

```tsx
const App = () => {
  if (process.env.MODERN_TARGET === 'browser') {
    const safeGet = require('lodash.get');
    const o = { a: 'x' };
    console.log(safeGet(o, 'a'));
    // or some sdk only run in browser.
  };
};

export default App;
```

Unlike `context.isBrowser`, this env var removes useless code after production bundle.

`context.isBrowser` can be used in most cases to differentiate logic at runtime.

But in some cases, such as files with side effects, or logic outside the component, you need to do it through this environment variables：

```ts
logDevice(window.innerHeight);
```

We also recommend using environment variables instead of the runtime, as this will make the application package size

## Types of changes

## Checklist

### DIFF
--- a/packages/runtime/plugin-runtime/src/ssr/cli/index.ts
+++ b/packages/runtime/plugin-runtime/src/ssr/cli/index.ts
@@ -60,7 +60,7 @@ export default (): CliPlugin<AppTools> => ({
             },
           },
           tools: {
-            webpackChain: (chain, { name, CHAIN_ID }) => {
+            webpackChain: (chain, { name, isServer, CHAIN_ID }) => {
               const userConfig = api.useResolvedConfigContext();
 
               if (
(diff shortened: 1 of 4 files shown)


packages/runtime/plugin-runtime/src/ssr/cli/index.ts
```diff
@@ -76,6 +76,26 @@ export default (): CliPlugin<AppTools> => ({
                     { filename: LOADABLE_STATS_FILE },
                   ]);
               }
+
+              // add environment variables to determine the node/browser
+              const prefix = `${
+                appContext.metaName.split(/[-_]/)[0]
+              }_`.toUpperCase();
+              const jupiterVars = {
```
chenjiahan, line 84: jupiterVars -> modernVars

packages/toolkit/main-doc/zh/guides/basic-features/env-vars.md
```diff
@@ -23,6 +23,36 @@ Modern.js 提供了对环境变量的支持，包含内置的环境变量和自
...
+    console.log(window.innerHeight);
+  };
+};
+```
+
+打包后可以看到 SSR 产物和 CSR 产物如下：
+
+```js title="dist/bundles/main.js"
+function App() {
+  if (false) {}
```
xuchaobei, line 42: if 语句会被移除掉吧？

xuchaobei (changes requested):

packages/runtime/plugin-runtime/src/ssr/cli/index.ts
zllkjc: done

packages/toolkit/main-doc/zh/guides/basic-features/env-vars.md
zllkjc: 嗯嗯，我补充了一个 note，这里是开发环境的打包产物。

xuchaobei (approved):

PR #2308, opened 2022-12-07
fix(builder): avoid inline files into html in dev mode

jkzing:
# PR Details

Keep consistent with the same behaviour in modern.js v1, according to code here https://github.com/modern-js-dev/modern.js/blob/main/packages/cli/webpack/src/config/client.ts#L86.

In case the inlined webpack runtime could cause the HMR hash not get updated during watch and compilation, which causes all past HRM requests to be played after a refresh, like this one:

CleanShot 2022-12-07 at 14 57 49@2x

## Types of changes

## Checklist


packages/builder/builder-webpack-provider/src/plugins/inlineChunk.ts
```diff
@@ -6,10 +6,10 @@ export const PluginInlineChunk = (): BuilderPlugin => ({
   name: 'builder-plugin-inline-chunk',
 
   setup(api) {
-    api.modifyWebpackChain(async (chain, { target, CHAIN_ID }) => {
+    api.modifyWebpackChain(async (chain, { target, CHAIN_ID, isProd }) => {
```
chenjiahan, line 9: disableInlineRuntimeChunk 的文档也可以改一下哈，添加「在生产环境下默认内联」

packages/builder/builder-webpack-provider/src/plugins/inlineChunk.ts
```diff
@@ -6,10 +6,10 @@ export const PluginInlineChunk = (): BuilderPlugin => ({
...
       const config = api.getNormalizedConfig();
 
-      if (isHtmlDisabled(config, target)) {
+      if (isHtmlDisabled(config, target) || !isProd) {
```
chenjiahan, line 12: 话说 enableInlineScripts 和 enableInlineStyles 也只在生产环境下生效吗，感觉让用户来决定比较好，这样能保持开发和生产环境一致

chenjiahan: soga，保持一致也没问题~ 那 enableInlineScripts 和 enableInlineStyles 的文档也得标注下，只在生产环境下生效

chenjiahan (approved):

PR #2708, opened 2023-01-17
feat(rspack-provider): rem plugin support

9aoy:
## Description

- reuse AutoSetRootFontSizePlugin
- support rem plugin in rspack
- support use chain.plugin in modifyBundlerChain
- different between rspack rem plugin and webpack rem plugin  （why not reuse）
   - STYLUS not support in rspack
   - not need import pxToRem in rspack （already integrated in @rspack/postcss-loader） 
   - HTMLPlugin is not same

## Types of changes

- [ ] Bug fix
- [ ] Refactoring
- [ ] Breaking change

## Checklist


packages/builder/builder-rspack-provider/src/plugins/rem.ts
```diff
@@ -0,0 +1,78 @@
+import _ from '@modern-js/utils/lodash';
```
chenjiahan, line 1: 补个 changeset

chenjiahan (approved):

PR #2712, opened 2023-01-18
feat: modern-js switch on rspack 

GiveMe-A-Name:
## Description

feat: using rspack to compiler modern-js project

## Types of changes

- [ ] Bug fix
- [ ] Refactoring
- [ ] Breaking change

## Checklist


packages/builder/builder-rspack-provider/src/types/rspack.ts
```diff
@@ -1,6 +1,9 @@
-import type { Compiler, RspackOptions } from '@rspack/core';
+import type * as rspack from '@rspack/core';
 
-export { Compiler };
+type Compiler = rspack.Compiler;
+type RspackOptions = rspack.RspackOptions;
+
+export type { Compiler, rspack };
```
chenjiahan, line 6: Makybe the `rspack` type should be `Rspack` (In PascalCase), the `webpack` type is a special case.

packages/builder/builder-shared/src/types/bundlerConfig.ts
```diff
@@ -22,6 +22,45 @@ type SplitChunks = Configuration extends {
   : never;
 
 type WebpackOptimization = NonNullable<Configuration['optimization']>;
+type WebpackResolve = NonNullable<Configuration['resolve']>;
+type WebpackOutput = NonNullable<Configuration['output']>;
+
+// fork from the @rspack/core
```
chenjiahan, line 28: Why fork this type

packages/cli/plugin-bff/src/cli.ts
```diff
@@ -16,7 +16,7 @@ import { registerModernRuntimePath } from './helper';
 const DEFAULT_API_PREFIX = '/api';
 const TS_CONFIG_FILENAME = 'tsconfig.json';
 
-export default (): CliPlugin<AppTools> => ({
+export default (): CliPlugin<AppTools<'webpack'>> => ({
```
chenjiahan, line 19: 'webpack' should be a default value, users don't need to explicitly declare it.

packages/solutions/app-tools/src/index.ts
```diff
@@ -110,7 +110,16 @@ export const buildCommand = async (
   }
 };
 
-export default (): CliPlugin<AppTools> => ({
+interface AppToolsOptions {
+  /** Specify the use what kind of bundler to compiler, default: `webpack` */
+  bundler?: 'experiment-rspack' | 'webpack';
```
chenjiahan, line 115: `experiment-rspack` -> `experimental-rspack`

packages/solutions/app-tools/package.json
```diff
@@ -75,6 +75,7 @@
     "@modern-js/builder-plugin-node-polyfill": "workspace:*",
     "@modern-js/builder-shared": "workspace:*",
     "@modern-js/builder-webpack-provider": "workspace:*",
+    "@modern-js/builder-rspack-provider": "workspace:*",
```
chenjiahan, line 78: `@modern-js/builder-rspack-provider` should not be declared in the `dependencies` field.

packages/cli/plugin-bff/src/cli.ts
GiveMe-A-Name: The AppTools CliPlugin has 3 types temporarily: `AppTools<'webpack'>` 、 `AppTools<'rspack'>` 、 `AppTools<undefined>`. The default type is `AppTools<undefined>`, and the CliPlugin can use SharedTypes in defines. This is means the CliPlugin not only use in the base of webpack but alse use in rspack base

packages/builder/builder-shared/src/types/bundlerConfig.ts
GiveMe-A-Name: 1. We need the Rspack type in `bulider-shared` to define some config or type
2. the `@rspack/core`  bundler size is so big that we can't  integration into `builder-shared`

so I fork the type temporarily

packages/solutions/app-tools/package.json
GiveMe-A-Name: done

packages/solutions/app-tools/src/index.ts
GiveMe-A-Name: done

packages/builder/builder-rspack-provider/src/types/rspack.ts
GiveMe-A-Name: done

packages/cli/plugin-bff/src/cli.ts
chenjiahan: It sounds like a breaking change for Plugin developers.

GiveMe-A-Name: Or we can make the `CliPlugin<'webpack'>` default? After all, `Rspack` is in experimental status. 

But with the development of the `Rspack`, It's  necessary to distinguish between 'webpack' Plugin and 'rspack' Plugin. 
When user is not specify the CliPlugin type, is should be shared CliPlugin.

chenjiahan: Yes, we should make the `CliPlugin<'webpack'>` as a default value.

GiveMe-A-Name: done

chenjiahan (approved):

9aoy (approved):

PR #2747, opened 2023-01-31
feat: add rspack minimize plugin

9aoy:
## Description

add rspack minimize plugin:
- support performance.removeConsole / output.disableMinimize
- not support  tools.minifyCss （cssNano） / tools.terser （terserOptions） /   output.legalComments

## Types of changes

- [ ] Bug fix
- [ ] Refactoring
- [ ] Breaking change

## Checklist


packages/builder/builder-rspack-provider/src/config/defaults.ts
```diff
@@ -29,7 +29,7 @@ export const createDefaultConfig = () =>
       // profile: false,
```
chenjiahan, line 29: 补个 changeset

chenjiahan (approved):

PR #2766, opened 2023-02-01
feat(core): support read local modern config

chenjiahan:
## Description

Supports creating `modern.config.local.ts` files to override `modern.config.ts` configurations.

```ts
// modern.config.ts
import { defineConfig } from '@modern-js/app-tools';
export default defineConfig({
  server: {
    port: 3000,
  },
});
```

```ts
// modern.config.local.ts
import { defineConfig } from '@modern-js/app-tools';
export default defineConfig({
  server: {
    port: 3001,
  },
});
```

The configuration in the `modern.config.local.ts` file will be deep merged with the configuration in `modern.config.ts` and override the configurations in `modern.config.ts`, so `server.port` will be is overridden by `3001`.

## Types of changes

- [ ] Bug fix
- [ ] Refactoring
- [ ] Breaking change

## Checklist


packages/cli/core/src/config/createLoadedConfig.ts
```diff
@@ -20,26 +32,84 @@ export const assignPkgConfig = (
     return undefined;
   });
 
+/**
+ * A modern config can export a function or an object
+ * If it's a function, it will be called and return a config object
+ */
+async function getConfigObject(config?: ConfigParams) {
+  if (typeof config === 'function') {
+    return (await config(0)) || {};
```
chenjiahan, line 41: The `config(0)` is an incorrect implementation, I will make another PR to fix it.

GiveMe-A-Name (approved):

PR #2818, opened 2023-02-08
feat(builder): reuse bundle analyzer plugin between rspack and webpack provider

9aoy:
## Types of changes

- [ ] Bug fix
- [ ] Refactoring
- [ ] Breaking change

## Checklist


packages/builder/builder-rspack-provider/src/shared/plugin.ts
```diff
@@ -36,4 +36,5 @@ export const applyDefaultPlugins = (plugins: Plugins) =>
     plugins.yaml(),
     plugins.splitChunks(),
     plugins.startUrl(),
+    plugins.bundleAnalyzer(),
```
chenjiahan, line 39: changeset 加一个

PR #2823, opened 2023-02-08
fix: load microApp can not match it's route

nyqykk:
## Description
fix: load microApp can not match it's route
fix: server.baseUrl is in preference to router.basename

custom route can not get basename correctly,so pass basename as props to the root component.
useMatchs hook return the pattern of the route,if not replace the redundant path,the basename will get a wrong value.
example: \<Route path='/sub/\* /\>,the matchs will get /sub/*,but the right basename is /sub

## Types of changes

- [x] Bug fix
- [ ] Refactoring
- [ ] Breaking change

## Checklist


.changeset/cold-snakes-invent.md
```diff
@@ -0,0 +1,9 @@
+---
+'@modern-js/plugin-garfish': patch
+---
+
+fix: load microApp can not match it's route
+fix: server.baseUrl is in preference to router.basename
+
+修复微前端模式下约定式路由和自控式路由子应用无法正确加载问题
```
zhoushaw, line 8: 前缀增加一下

zhoushaw (approved):

PR #2890, opened 2023-02-14
fix: use entryDir to cal doc file

clChenLiang:
## Description

 修复 document 文件获取的 bug

## Types of changes

- [ x ] Bug fix
- [ ] Refactoring
- [ ] Breaking change

## Checklist


packages/solutions/app-tools/src/analyze/getBundleEntry.ts
```diff
@@ -64,6 +64,7 @@ export const getBundleEntry = (
           ? {
               entryName: name,
               entry: ensureAbsolutePath(appDirectory, value),
+              entryDir: path.dirname(ensureAbsolutePath(appDirectory, value)),
```
jkzing, line 67: set this value to `absoluteEntryDir` could be enough

jkzing: otherwise `entryDir` could be empty in nested routes analyzed entries

clChenLiang: fixed this with make the getBundleEntry calculate the absoluteEntryDir too.
The problem in nested routes entries not have absoluteEntryDir is because in nested route:
image

yimingjfe (approved):

PR #2908, opened 2023-02-16
fix(plugin-storybook): remove dist dir when build

targeral:
## Description

remove dist dir when run `modern build`

## Types of changes

- [x] Bug fix
- [ ] Refactoring
- [ ] Breaking change

## Checklist

### DIFF
--- a/packages/cli/plugin-storybook/src/index.ts
+++ b/packages/cli/plugin-storybook/src/index.ts
@@ -1,3 +1,4 @@
+import path from 'path';
 import { createRuntimeExportsUtils } from '@modern-js/utils';
 import type { CliPlugin, ModuleTools } from '@modern-js/module-tools';
 import { defaultStories, appToolsStories } from './constants/stores';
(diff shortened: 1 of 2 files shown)


packages/cli/plugin-storybook/src/index.ts
```diff
@@ -39,6 +40,19 @@ export default (pluginOption: PluginOptions = {}): CliPlugin<ModuleTools> => ({
       return config;
     },
 
+    beforeBuild: async () => {
+      const { fs } = await import('@modern-js/utils');
+      const { STORYBOOK_DIST_DIR_NAME } = await import('./features/constants');
+      const { appDirectory } = api.useAppContext();
+      const storybookDistPath = path.join(
+        appDirectory,
+        'dist',
```
chenjiahan, line 49: We can use`appContext.distDirectory`

targeral: How is `distDirectory` calculated? It may not have much to do with Module Tools

targeral: This output path is currently temporary and an api should be designed later.

chenjiahan: You are right, ignore this comment.

chenjiahan (approved):

PR #2935, opened 2023-02-20
feat: add noDevTools Hook and change dev log

targeral:
## Description

add noDevTools hook, fix dev log content and add chinese and english content

image

## Types of changes

- [x] Bug fix
- [ ] Refactoring
- [ ] Breaking change

## Checklist

### DIFF
--- a/packages/solutions/module-tools/src/locale/en.ts
+++ b/packages/solutions/module-tools/src/locale/en.ts
@@ -1,15 +1,25 @@
+import { chalk } from '@modern-js/utils';
+
+const noDevTools = `
+There are no DevTools available, you can learn about them and choose to use them by following the link options.
+- ${chalk.underline(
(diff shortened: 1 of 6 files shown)


packages/solutions/module-tools/src/locale/en.ts
```diff
@@ -25,4 +35,9 @@ export const EN_LOCALE = {
       lang: 'set new command language(en or zh)',
     },
   },
+  log: {
+    dev: {
+      no_devtools: noDevTools,
```
chenjiahan, line 40: image

This file uses a mix of camelCase and underscore styles, maybe we should unify the naming.

targeral: update

chenjiahan (approved):

PR #3004, opened 2023-02-24
fix: disable swc react refresh when ssr mode.|

GiveMe-A-Name:
## Description

react refresh should use in web broswer,
it seems inject a RefreshReg function call.

so we need disable it when ssr mode

## Types of changes

- [x] Bug fix
- [ ] Refactoring
- [ ] Breaking change

## Checklist

### DIFF
--- a/packages/builder/plugin-swc/src/plugin.ts
+++ b/packages/builder/plugin-swc/src/plugin.ts
@@ -1,5 +1,5 @@
 import path from 'path';
-import { Compiler, Compilation } from 'webpack';
+import type { Compiler, Compilation } from 'webpack';
 import type { BuilderPluginAPI } from '@modern-js/builder-webpack-provider';
 import {
   mergeRegex,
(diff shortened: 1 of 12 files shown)


packages/builder/plugin-swc/src/plugin.ts
```diff
@@ -49,7 +49,10 @@ export const builderPluginSwc = (
         jsc: {
           transform: {
             react: {
-              refresh: env === 'development' && builderConfig.dev.hmr,
+              refresh:
```
chenjiahan, line 52: Maybe we can reuse this method:

image

GiveMe-A-Name: Yeap, Right.
Hold on, I am modifying.

GiveMe-A-Name: It seems many `isUsingHRM` in difference packages, e.g: `@modern-js/builder-rspack-provider`, `@modern-js/builder-webpack-provider"`.

It may be better that If we put the function into `@modern-js/builder-shared`, 
then  we let these packages, e.g `@modern-js/builder-rspack-provider`, `@modern-js/builder-plugin-swc` to import the `@modern-js/builder-shared`

chenjiahan: Yes, `isUsingHMR` should be moved to the `@modern-js/builder-shared` package.

chenjiahan (dismissed):

JSerFeng (approved):

PR #3084, opened 2023-03-08
feat(app-tools): the modern.js ssr loader support rspack

GiveMe-A-Name:
## Description

improvement: the modern.js ssr loader support rspack            

the `rspack` is not support virtualModule
so we write the `ssr` loader combinedModule in filesystem;  
then we load it from disk.                          

It is temporary.Once `rspack` support it, we will use   
virtualModule to implement it.

## Types of changes

- [ ] Bug fix
- [ ] Refactoring
- [ ] Breaking change

## Checklist

### DIFF
--- a/packages/solutions/app-tools/src/analyze/generateCode.ts
+++ b/packages/solutions/app-tools/src/analyze/generateCode.ts
@@ -29,7 +29,11 @@ import {
   ENTRY_POINT_FILE_NAME,
   ENTRY_BOOTSTRAP_FILE_NAME,
 } from './constants';
-import { getDefaultImports, getServerLoadersFile } from './utils';
+import {
+  getDefaultImports,
+  getServerLoadersFile,
+  getServerCombinedModueFile,
+} from './utils';
 import { walk } from './nestedRoutes';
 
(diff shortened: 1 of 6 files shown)


packages/solutions/app-tools/src/analyze/generateCode.ts
```diff
@@ -204,6 +208,21 @@ export const generateCode = async (
...
+          config,
+          appContext,
+        );
+        if (serverLoaderCombined) {
+          const serverLoaderFile = getServerCombinedModueFile(
+            internalDirectory,
+            entryName,
+          );
+          await fs.ensureFile(serverLoaderFile);
+          await fs.writeFile(serverLoaderFile, serverLoaderCombined);
```
chenjiahan, line 223: We can use `fs.outputFile` instead of `fs.ensureFile` + `fs.writeFile`

GiveMe-A-Name: Got it

GiveMe-A-Name: Done~

chenjiahan (approved):

PR #3140, opened 2023-03-14
feat: rspack-provider support tools.babel

GiveMe-A-Name:
## Description

In some times, user maybe need add custom babel plugin or babel presets. 
`rspack-provider ` need support it

### UNSAFE

1. should we need add core-js for every entries in rspack?

2. the rspack is not support includes is a object. So It is a unsafe action.
```ts
  // compile all folders in app directory, exclude node_modules
  rule.include.add(context.rootPath);

  // it will let user can't include the package in `node_modules`.
  rule.exclude.add(/node_modules/);

  // ....
``` 

## Types of changes

- [ ] Bug fix
- [ ] Refactoring
- [ ] Breaking change

## Checklist


packages/builder/builder-shared/package.json
```diff
@@ -67,6 +68,7 @@
     }
   },
   "dependencies": {
+    "@modern-js/babel-preset-app": "workspace:*",
```
chenjiahan, line 71: Most Rspack projects do not need babel, if we declare the "@modern-js/babel-preset-app" dependency in `builder-shared` by default, the install size will be bigger.

packages/builder/builder-rspack-provider/src/plugins/babel.ts
```diff
@@ -0,0 +1,103 @@
...
+      rule
+        .test(mergeRegex(JS_REGEX, TS_REGEX))
+        .use(CHAIN_ID.USE.BABEL)
+        .loader(getSharedPkgCompiledPath('babel-loader'))
+        .options(babelOptions);
+    });
+  },
+});
+
+export function applyScriptCondition({
```
9aoy, line 78: 这个方法可以抽到 shared 里吧？

packages/builder/builder-webpack-provider/src/plugins/babel.ts
```diff
@@ -170,7 +164,7 @@ export const builderPluginBabel = (): BuilderPlugin => ({
         rule
           .test(useTsLoader ? JS_REGEX : mergeRegex(JS_REGEX, TS_REGEX))
           .use(CHAIN_ID.USE.BABEL)
-          .loader(getCompiledPath('babel-loader'))
+          .loader(getSharedPkgCompiledPath('babel-loader'))
```
9aoy, line 167: prebundle 的脚本不需要改么？  webpack-provider 里的 babel-loader bundle 也要移除吧

packages/builder/builder-rspack-provider/src/plugins/babel.ts
```diff
@@ -0,0 +1,103 @@
...
+        context: api.context,
+        includes,
+        excludes,
+      });
+
+      rule
+        .test(mergeRegex(JS_REGEX, TS_REGEX))
+        .use(CHAIN_ID.USE.BABEL)
+        .loader(getSharedPkgCompiledPath('babel-loader'))
+        .options(babelOptions);
```
9aoy, line 73: e2e 测试补充一个吧

packages/builder/builder-webpack-provider/src/plugins/babel.ts
GiveMe-A-Name: webpack-provider 里的 babel-loader bundle 移除了

packages/builder/builder-rspack-provider/src/plugins/babel.ts
GiveMe-A-Name: Got it

packages/builder/builder-shared/package.json
GiveMe-A-Name: Maybe in long time, the app-tools would default dependent the `webpack-provider`, It is mean we already dependent the `@modern-js/babel-preset-app` by default. 

That's a problem exactly, but it 's not a problem now.
We should resolve this when we switch the bundler engine to rspack by default

packages/builder/builder-rspack-provider/src/plugins/babel.ts
GiveMe-A-Name: done

packages/builder/builder-shared/package.json
GiveMe-A-Name: emm. Maybe we need think carefully about this problem.

9aoy (dismissed):

packages/builder/builder-rspack-provider/src/plugins/babel.ts
GiveMe-A-Name: Done~

packages/builder/builder-shared/package.json
GiveMe-A-Name: Updated it.`@modern-js/babel-preset-app` would be a dev dependency in `builder-shared`.

but the `babel-loader` still in `builder-shared`, There are two main reasons follow: 
- `babel-loader` only 242 kb size.
- The preBuild dependency does not load well dynamically

chenjiahan (approved):

9aoy (approved):

PR #3157, opened 2023-03-15
feat(builder): improve vendor split chunk rule

Asuka109:
## Types of changes

- [ ] Bug fix
- [ ] Refactoring
- [ ] Breaking change

## Checklist


packages/builder/builder/src/plugins/splitChunks.ts
```diff
@@ -92,15 +100,18 @@ async function splitByExperience(
     packageRegExps.antd = createDependenciesRegExp('antd');
   }
   if (isPackageInstalled('@arco-design/web-react', rootPath)) {
-    packageRegExps.arco = createDependenciesRegExp('arco-design');
+    packageRegExps.arco = createDependenciesRegExp(/@?arco-design/);
   }
   if (isPackageInstalled('@douyinfe/semi-ui', rootPath)) {
-    packageRegExps.semi = createDependenciesRegExp('semi-ui');
+    packageRegExps.semi = createDependenciesRegExp(
+      /@(ies|douyinfe)[\\/]semi-.*/,
```
chenjiahan, line 107: How about `/@(ies|douyinfe)[\\/]semi-/`, the result should be same.

Asuka109: This expression will be inserted into the template string, which includes path separator. So i think we should take a wildcard here.

chenjiahan (approved):

PR #3195, opened 2023-03-20
feat(app-tools): support add node-polyfill when use rspack

9aoy:
## Types of changes

- [ ] Bug fix
- [ ] Refactoring
- [ ] Breaking change

## Checklist


packages/solutions/app-tools/package.json
```diff
@@ -86,6 +86,7 @@
     "@modern-js/types": "workspace:*",
     "@modern-js/upgrade": "workspace:*",
     "@modern-js/utils": "workspace:*",
+    "@rspack/plugin-node-polyfill": "^0.1.1",
```
chenjiahan, line 89: `app-tools` should not depend on this package by default, maybe it can be moved to the `builder-rspack-provider`.

9aoy: In webpack-provider, it's an independent plugin which named `@modern-js/builder-plugin-node-polyfill`. how about I move it into builder-plugin-node-polyfill, and export as  builderRspackPluginNodePolyfill ?

chenjiahan: Why not just make the `@modern-js/builder-plugin-node-polyfill` to be compatible with Rspack?

9aoy: Rspack not support ProvidePlugin and `resolve.fallback`.  And in `@rspack/plugin-node-polyfill`, It's a limited node-polyfill support which not support global polyfill such as Buffer and process.

chenjiahan (approved):

PR #3206, opened 2023-03-21
feat(builder): security.checkSyntax support exclude configuration

GiveMe-A-Name:
## Description

feat(builder-webpack-proivder): security.checkSyntax support exclude configuration 

## Types of changes

- [ ] Bug fix
- [ ] Refactoring
- [ ] Breaking change

## Checklist


packages/builder/builder-webpack-provider/src/webpackPlugins/CheckSyntaxPlugin/helpers/utils.ts
```diff
@@ -0,0 +1,14 @@
...
+  path: string,
+  exclude?: CheckSyntaxExclude,
+) {
+  if (!exclude) {
+    return false;
+  }
+
+  const excludes = Array.isArray(exclude) ? exclude : [exclude];
+
+  return excludes.some(reg => new RegExp(reg).test(path));
```
chenjiahan, line 13: This behavior is different from the `exclude` of webpack rules, we have two options:

1. Only support `RegExp | RegExp[]`
2. Make the behavior consistent with webpack exclude.

GiveMe-A-Name: Done~

chenjiahan (approved):

PR #3270, opened 2023-03-28
fix(builder): `normalizeConfig` will merge arrays when filling default values for config

Asuka109:
## Description

Add new util function`defaultsToBuilderConfig`, which is used to replace `mergeBuilderConfig` in filling the default config.

## Related Issue

The new option `source.transformImports` can be an array. The `mergeBuilderConfig` function will be executed twice and the default values will be filled in two times:

```ts
const config = api.getNormalizedConfig().source.transformImports;
//        ^? EXPECT: ['a', 'b']
//           ACTUAL: ['a', 'b', 'a', 'b']
```

## Types of changes

- [x] Bug fix
- [ ] Refactoring
- [ ] Breaking change

## Checklist

### DIFF
--- a/packages/builder/builder-shared/src/mergeBuilderConfig.ts
+++ b/packages/builder/builder-shared/src/mergeBuilderConfig.ts
@@ -1,29 +1,37 @@
 import _ from '@modern-js/utils/lodash';
 
+export const defaultsToBuilderConfig = <T>(...configs: T[]): T =>
+  _.mergeWith({}, ...configs, (target: unknown, source: unknown) => {
+    const pair = [target, source];
+
+    // Use the defined one.
+    if (pair.some(_.isUndefined)) {
+      return target ?? source;
+    }
+
+    // Handle on each properties.
+    if (_.every(pair, _.isPlainObject)) {
+      return undefined;
+    }
+
+    return target;
+  });
+
 export const mergeBuilderConfig = <T>(...configs: T[]): T =>
-  _.mergeWith(
-    {},
-    ...configs,
-    (target: unknown, source: unknown, key: string) => {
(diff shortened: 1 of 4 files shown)


packages/builder/builder-shared/src/mergeBuilderConfig.ts
```diff
@@ -1,29 +1,37 @@
...
-  _.mergeWith(
-    {},
-    ...configs,
-    (target: unknown, source: unknown, key: string) => {
-      const pair = [target, source];
-      if (pair.some(_.isUndefined)) {
-        // fallback to lodash default merge behavior
-        return undefined;
-      }
+export const defaultsToBuilderConfig = <T>(...configs: T[]): T =>
```
chenjiahan, line 3: 要不叫 applyDefaultConfig?

packages/builder/builder-webpack-provider/src/config/normalize.ts
```diff
@@ -15,7 +18,7 @@ const createNormalizedDefaultConfig = () =>
  * 3. Meaningful and can be filled by constant value.
  */
 export const normalizeConfig = (config: BuilderConfig): NormalizedConfig =>
-  mergeBuilderConfig<NormalizedConfig>(
+  defaultsToBuilderConfig<NormalizedConfig>(
```
chenjiahan, line 21: 正常来说是 userConfig 在前，defaultConfig 在后，后者向前者填充。

packages/builder/builder-shared/src/mergeBuilderConfig.ts
chenjiahan, line 3: 单测++

PR #3288, opened 2023-03-29
feat(plugin-image-compress): compatible array config

Asuka109:
## Description

https://github.com/web-infra-dev/modern.js/blob/dc44482df996be9c221b84141ea92f24ffac720b/packages/builder/plugin-image-compress/tests/index.test.ts#L32-L86

## Types of changes

- [ ] Bug fix
- [ ] Refactoring
- [ ] Breaking change

## Checklist

### DIFF
--- a/packages/builder/plugin-image-compress/src/index.ts
+++ b/packages/builder/plugin-image-compress/src/index.ts
@@ -1,4 +1,5 @@
 import type { BuilderPlugin } from '@modern-js/builder-webpack-provider/types';
+import assert from 'assert';
 import { ModernJsImageMinimizerPlugin } from './minimizer';
 import { withDefaultOptions } from './shared/utils';
 import { Codecs, Options } from './types';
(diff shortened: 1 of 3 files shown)


packages/builder/plugin-image-compress/src/index.ts
```diff
@@ -1,19 +1,46 @@
...
+const castOptions = (args: (Options | Options[])[]): Options[] => {
+  const head = args[0];
+  // expect [['png', { use: 'jpeg' }]]
+  if (Array.isArray(head)) {
+    return head;
+  }
+  // expect ['png', { use: 'jpeg' }]
+  const ret: Options[] = [];
+  for (const arg of args) {
+    assert(!Array.isArray(arg));
```
chenjiahan, line 24: why do we need to assert the arg

Asuka109: Just for typing, not necessary. TypeScript can't distinguish between the two subtypes of the union type.

PR #3367, opened 2023-04-06
fix: the defineLegacyConfig should set autoLoadPlugins = true by default

GiveMe-A-Name:
### `defineLegacyConfig` Bug

We set a `legacy` symbol to tag the action when user use `defineLegacyConfig` API, then set `autoLoadPlugins` = true in "config" hooks.
But the "config" hook run before loadPlugin.
image
So we set `autoLoadPlugins` config before all;

## Types of changes

- [x] Bug fix
- [ ] Refactoring
- [ ] Breaking change

## Checklist


jkzing (approved):
LGTM👍

PR #3405, opened 2023-04-11
perf(plugin): reduce bundle size

chenjiahan:
## Description

The `@modern-js/plugin` package is used both in Node.js and browser, so we should make the bundle size smaller to reduce the size of Modern.js application.

This PR have adjusted the build config of `@modern-js/plugin` to reduce duplicated helpers, the `dist/esm/index.js` will be bundled rather than bundleless.

Also adjusted some syntaxes to reduce the generated code, such as `async` and `for of`.

The minified size of `@modern-js/plugin` is reduced from `34.67kb` to `27.4kb`.

### Before

Screen Shot 2023-04-11 at 17 46 44

### After

Screen Shot 2023-04-11 at 19 33 04

## Types of changes

- [ ] Bug fix
- [ ] Refactoring
- [ ] Breaking change

## Checklist


packages/toolkit/plugin/src/manager/async.ts
```diff
@@ -146,10 +146,10 @@ export const createAsyncManager = <
           addPlugin(createPlugin(plugin.setup, plugin));
         }
         // unknown plugin
-        else {
+        else if (process.env.NODE_ENV !== 'production') {
```
targeral, line 149: Why add this judgment logic?

PR #3436, opened 2023-04-13
chore(builder): clean up unnecessary e2e files

chenjiahan:
## Summary

Clean up unnecessary e2e files, so we can focus on the test cases rather than unused files.

Most e2e cases only need to provide a `index.test.ts` and some source files, the `package.json` and `*.md` is not required.

## Checklist


tests/e2e/builder/cases/basic/build.test.ts
Asuka109, line 1: @chenjiahan 这个也要删掉嘛

chenjiahan: 这个 case 好像没有什么实际的作用。。

PR #3440, opened 2023-04-13
fix(app-tools): inline-loader match path error in windows

GiveMe-A-Name:
### <samp>🤖 Generated by Copilot at da2b838</samp>

This pull request fixes a bug that prevented the inline-loader from working properly in Windows environments. It updates the `@modern-js/app-tools` package with a patch version and a changelog entry, and changes the path matching logic in `adapterSSR.ts`.

### <samp>🤖 Generated by Copilot at da2b838</samp>

*  Fix inline-loader path matching for Windows in `adapterSSR.ts` ([link](https://github.com/web-infra-dev/modern.js/pull/3440/files?diff=unified&w=0#diff-d019db18fbb5169d98250c217097e90337f356e26121269d691b402495334186L237-R242), [link](https://github.com/web-infra-dev/modern.js/pull/3440/files?diff=unified&w=0#diff-716000dc174aabf493a50b869103e7263590d08e50935c53bbb671391dd3106cR1-R6))

## Checklist

### DIFF
--- a/packages/solutions/app-tools/src/builder/shared/builderPlugins/adapterSSR.ts
+++ b/packages/solutions/app-tools/src/builder/shared/builderPlugins/adapterSSR.ts
@@ -234,6 +234,8 @@ function applySSRDataLoader<B extends Bundler>(
 
   const { entriesDir = './src' } = normalizedConfig.source;
 
+  // Both match in Windows and Unix(Linux, Mac)
+  // docs: https://github.com/webpack/webpack/issues/2073
   const absolutePath = path
     .resolve(appDirectory, entriesDir)
     .split(path.sep)
--- a/packages/solutions/app-tools/src/builder/shared/builderPlugins/adapterSSR.ts
+++ b/packages/solutions/app-tools/src/builder/shared/builderPlugins/adapterSSR.ts
@@ -234,7 +234,10 @@ function applySSRDataLoader<B extends Bundler>(
 
   const { entriesDir = './src' } = normalizedConfig.source;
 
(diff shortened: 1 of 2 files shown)


packages/solutions/app-tools/src/builder/shared/builderPlugins/adapterSSR.ts
```diff
@@ -234,7 +234,12 @@ function applySSRDataLoader<B extends Bundler>(
 
   const { entriesDir = './src' } = normalizedConfig.source;
 
-  const absolutePath = path.resolve(appDirectory, entriesDir);
+  // Both match in Windows and Unix(Linux, Mac)
+  // docs: https://github.com/webpack/webpack/issues/2073
+  const absolutePath = path
+    .resolve(appDirectory, entriesDir)
+    .split(path.sep)
+    .join('(\\\\|/)');
```
chenjiahan, line 242: How about using the `normalizeToPosixPath` util

GiveMe-A-Name: In template, we will product a loader path like this:
```js
import loader_0 from 'C:/a/b/c.loader.ts'
```
We need match this path use data-loader.
If we transform the path by using `normalizeToPosixPath`, also need  transform the template path. Like this:
```js
import loader_0 from '/C/a/b/c.loader.ts'
```

But if we use /C/a/b/c.loader.ts to match real files, wouldn't the windows system be able to find the file?

chenjiahan:
The PR title needs to conform to the naming convention

yimingjfe (approved):

PR #3443, opened 2023-04-14
fix(builder-shared): The http://0.0.0.0:port can't visit in windows

GiveMe-A-Name:
### <samp>🤖 Generated by Copilot at 4911ea3</samp>

This pull request fixes a bug in the asset-prefix feature of the `@modern-js/builder-shared` package, which caused problems in Windows systems when using `0.0.0.0` as the hostname for the publicPath option in the webpack output configuration. It updates the `getPublicPath` function, the `index.test.ts` file, and the `.changeset` file accordingly.

### <samp>🤖 Generated by Copilot at 4911ea3</samp>

*  Fix the issue of using `0.0.0.0` as the hostname for the publicPath option in the webpack output configuration, which causes problems in Windows systems ([link](https://github.com/web-infra-dev/modern.js/pull/3443/files?diff=unified&w=0#diff-44fa071163a9ffcf5a02878088470b75c77096217375ce7c4d5ef6fd6d437efeR1-R6), [link](https://github.com/web-infra-dev/modern.js/pull/3443/files?diff=unified&w=0#diff-5071b6a55bc61e45b3db5266c355c86bb495f414f628f6aebaf5584a1db86911L85-R91))
  * Add a `.changeset` file with fix messages in English and Chinese, indicating a patch-level change for the `@modern-js/builder-shared` package ([link](https://github.com/web-infra-dev/modern.js/pull/3443/files?diff=unified&w=0#diff-44fa071163a9ffcf5a02878088470b75c77096217375ce7c4d5ef6fd6d437efeR1-R6))
  * Modify the `getPublicPath` function in `output.ts` to check if the hostname is not equal to `0.0.0.0`, and only set the publicPath to `//${hostname}:${port}/` if that is the case, otherwise leave it as an empty string ([link](https://github.com/web-infra-dev/modern.js/pull/3443/files?diff=unified&w=0#diff-5071b6a55bc61e45b3db5266c355c86bb495f414f628f6aebaf5584a1db86911L85-R91))
  * Add a code comment explaining the rationale for the change and referencing a related issue in another project ([link](https://github.com/web-infra-dev/modern.js/pull/3443/files?diff=unified&w=0#diff-5071b6a55bc61e45b3db5266c355c86bb495f414f628f6aebaf5584a1db86911L85-R91))
* Update the test file for the asset-prefix feature of the `@modern-js/builder-shared` package to reflect the change in the publicPath option ([link](https://github.com/web-infra-dev/modern.js/pull/3443/files?diff=unified&w=0#diff-1f176658b2041da3513c5609802ebb998a0320326340449f955df1f048555eabL4), [link](https://github.com/web-infra-dev/modern.js/pull/3443/files?diff=unified&w=0#diff-1f176658b2041da3513c5609802ebb998a0320326340449f955df1f048555eabL21-R20), [link](https://github.com/web-infra-dev/modern.js/pull/3443/files?diff=unified&w=0#diff-1f176658b2041da3513c5609802ebb998a0320326340449f955df1f048555eabL30-R29), [link](https://github.com/web-infra-dev/modern.js/pull/3443/files?diff=unified&w=0#diff-1f176658b2041da3513c5609802ebb998a0320326340449f955df1f048555eabL41-R40))
  * Remove an unused import statement of the `DEFAULT_DEV_HOST` constant from `@modern-js/utils` ([link](https://github.com/web-infra-dev/modern.js/pull/3443/files?diff=unified&w=0#diff-1f176658b2041da3513c5609802ebb998a0320326340449f955df1f048555eabL4))
  * Modify an assertion to check only for the `/static/js/` part of the publicPath in the HTML file, instead of the `//${DEFAULT_DEV_HOST}:3333` part ([link](https://github.com/web-infra-dev/modern.js/pull/3443/files?diff=unified&w=0#diff-1f176658b2041da3513c5609802ebb998a0320326340449f955df1f048555eabL21-R20))
  * Modify a variable assignment to set the expected publicPath to an empty string, instead of `//${DEFAULT_DEV_HOST}:3333` ([link](https://github.com/web-infra-dev/modern.js/pull/3443/files?diff=unified&w=0#diff-1f176658b2041da3513c5609802ebb998a0320326340449f955df1f048555eabL30-R29))
  * Modify a page navigation to use `http://localhost:3333` as the URL, instead of `http:${expected}` ([link](https://github.com/web-infra-dev/modern.js/pull/3443/files?diff=unified&w=0#diff-1f176658b2041da3513c5609802ebb998a0320326340449f955df1f048555eabL41-R40))

## Checklist


packages/builder/builder-shared/src/apply/output.ts
```diff
@@ -82,7 +82,13 @@ function getPublicPath({
   } else if (dev.assetPrefix === true) {
     const hostname = context.devServer?.hostname || DEFAULT_DEV_HOST;
     const port = context.devServer?.port || DEFAULT_PORT;
-    publicPath = `//${hostname}:${port}/`;
+    if (hostname !== DEFAULT_DEV_HOST) {
```
chenjiahan, line 85: Maybe we should revert this Pull Request: https://github.com/web-infra-dev/modern.js/pull/2975/files

0.0.0.0 should not be used as default dev host, it is not same as `localhost`:

> In the context of a Windows system, 0.0.0.0 and localhost are not equivalent.

> 0.0.0.0 is a special IP address that represents all available IP addresses on the local machine. When a server is configured to bind to 0.0.0.0, it will listen on all available network interfaces, including the loopback interface (127.0.0.1) and any physical network interfaces.

> localhost, on the other hand, is a hostname that resolves to the loopback IP address (127.0.0.1). It is commonly used to refer to the local machine and is often used as a default hostname for local web servers.

> So while both 0.0.0.0 and localhost can be used to refer to the local machine, they have different meanings and uses.

packages/builder/builder-shared/src/apply/output.ts
```diff
@@ -82,8 +82,19 @@ function getPublicPath({
...
+      // The http://0.0.0.0:port can't visit in windows, so we shouldn't set publicPath as `//0.0.0.0:${port}/`;
+      // Relative to docs:
+      // - https://github.com/quarkusio/quarkus/issues/12246
+      publicPath = `//${localHostname}:${port}/`;
+    } else {
+      publicPath = `//${hostname}:${port}/`;
+    }
   }
 
+  console.info('publicPath', publicPath);
```
9aoy, line 97: debug？

packages/builder/builder-shared/src/apply/output.ts
```diff
@@ -82,7 +82,16 @@ function getPublicPath({
   } else if (dev.assetPrefix === true) {
     const hostname = context.devServer?.hostname || DEFAULT_DEV_HOST;
     const port = context.devServer?.port || DEFAULT_PORT;
-    publicPath = `//${hostname}:${port}/`;
+    if (hostname === DEFAULT_DEV_HOST) {
+      const localHostname = `127.0.0.1`;
```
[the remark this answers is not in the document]
GiveMe-A-Name: Got it

chenjiahan (approved):

chenjiahan (approved):

Issue #3527, opened 2023-04-21
[RFC]: Modern.js Doc support edit this page on GitHub


chenjiahan:
### What problem does this feature solve?

Support "Edit this page on GitHub" link, making it easier to update the document.

![image](https://user-images.githubusercontent.com/7237365/233558584-92261bb8-36c1-428b-9c08-4c57c503a55c.png)

### What does the proposed API look like?

The text and repo link can be configured, so we can display "Edit this page on GitLab" or something else.

PR #3590, opened 2023-04-28
feat: modern.js doc support edit this page on github (#3527)

RyanHo97:
在这个版本中，添加了modern.js doc edit this page on github支持，通过themeConfig中的editLink配置repoUrl仓库地址还有text文本提示。
这个功能建议是由issues在 #3527 的讨论中提出的。

### DIFF
--- a/.changeset/swift-donuts-shout.md
+++ b/.changeset/swift-donuts-shout.md
@@ -1,0 +1,6 @@
+---
+"@modern-js/doc-core": patch
+---
+
+fix(doc-core): The 960px-1440px and 1440px center container widths do not include the width of the scroll bar to avoid horizontal scroll bars
+fix(doc-core): 960px-1440px和1440px版心容器宽度不包含滚动条宽度，避免出现横向滚动条
--- a/packages/cli/doc-core/src/theme-default/layout/DocLayout/index.module.scss
+++ b/packages/cli/doc-core/src/theme-default/layout/DocLayout/index.module.scss
@@ -36,7 +36,7 @@
 
   .content {
     margin-left: var(--modern-sidebar-width);
-    width: calc(100vw - var(--modern-sidebar-width));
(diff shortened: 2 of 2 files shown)


package.json
```diff
@@ -43,9 +43,9 @@
   },
   "engines": {
     "node": ">=14.17.6",
-    "pnpm": ">=7.12.2 <=7.29.1"
+    "pnpm": ">=7.12.2 <=8.2.0"
   },
-  "packageManager": "pnpm@7.29.1",
+  "packageManager": "pnpm@8.2.0",
```
chenjiahan, line 48: We should create a separate pull request to upgrade the pnpm version,  please downgrade the pnpm version in this PR.

RyanHo97: thanks,I want to improve this place tomorrow

packages/document/doc-tools-doc/modern.config.ts
```diff
@@ -50,6 +50,11 @@ export default defineConfig({
           label: 'English',
         },
       ],
+      editLink: {
+        repoUrl: 'https://github.com/test/blog',
+        text: 'Edit this page on GitHub',
+        pattern: '',
```
sanyuan0704, line 56: The `pattern` param is not neccessary because we can join  doc repo base url and relative page path and take the result as edit link url.

packages/cli/doc-core/src/theme-default/components/DocFooter/index.tsx
```diff
@@ -53,6 +59,9 @@ export function DocFooter() {
           ) : null}
         </div>
       </div>
+      <div className="flex flex-col">
+        <EditLink editLink={editLink} relativePagePath={''} />
```
sanyuan0704, line 63: The relativePagePath should be computed.

PR #3706, opened 2023-05-17
feat: support add custom tempdir for modern.js App

zllkjc:
## Summary

Modern.js will create tempDir for the real webpack entry before build. If there is multiple config in one app, they can not build together. Now we support a new params for `@modern-js/app-tools`:

```ts
import path from 'path';
import appTools, { defineConfig } from '@modern-js/app-tools';

export default defineConfig({
  output: {
    tempDir: path.join('node_modules', '.temp-dir'),
  },
  plugins: [appTools({})],
});

```

## Checklist

### DIFF
--- a/packages/solutions/app-tools/src/index.ts
+++ b/packages/solutions/app-tools/src/index.ts
@@ -115,7 +115,6 @@ export const buildCommand = async (
 export type AppToolsOptions = {
   /** Specify the use what kind of bundler to compiler, default: `webpack` */
   bundler?: 'experimental-rspack' | 'webpack';
-  tempDir?: string;
 };
 
 export default (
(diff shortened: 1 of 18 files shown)


packages/solutions/app-tools/src/index.ts
```diff
@@ -154,6 +154,19 @@ export default (
     i18n.changeLanguage({ locale });
 
     return {
+      async afterInit() {
```
chenjiahan, line 157: How about running the following codes in the body of `setup` function:

image

chenjiahan: Oh It doesn't seem to work, the AppContext has not been initialized when the setup is executed,

packages/solutions/app-tools/src/types/config/output.ts
```diff
@@ -14,6 +14,7 @@ export interface SharedOutputConfig extends BuilderSharedOutputConfig {
   ssg?: SSGConfig;
   splitRouteChunks?: boolean;
   disableNodePolyfill?: boolean;
+  tempDir?: string;
```
chenjiahan, line 17: The `output.tempDir` config should be documented.

packages/solutions/app-tools/src/index.ts
chenjiahan: How about renaming `afterInit` to `beforeConfig`, because there is no `init` hook

zllkjc: Okey, that's better~

chenjiahan: The `beforeConfig` should also be documented 😄

zllkjc: done~

chenjiahan (approved):

PR #3725, opened 2023-05-22
feat: support security.nonce for add nonce attribute on script tag

zllkjc:
## Summary

copilot:summary

## Details

copilot:walkthrough

## Checklist


packages/builder/builder-rspack-provider/src/types/config/security.ts
```diff
@@ -1,5 +1,16 @@
-// TODO
-// eslint-disable-next-line @typescript-eslint/no-empty-interface
-export interface SecurityConfig {}
+// Todo support security.sri configuration
+/**
+ * Currently, rspack does not support the security.sri configuration.
```
chenjiahan, line 3: @9aoy Can we support security.sri in Rspack?

packages/runtime/plugin-runtime/src/ssr/serverRender/utils.ts
```diff
@@ -25,3 +25,9 @@ export function getLoadableScripts(extractor: ChunkExtractor) {
       .join('')
   );
 }
+export function attributesToString(attributes: Record<string, any>) {
+  // 遍历属性并转化为字符串，只包含非 undefined 的属性
```
chenjiahan, line 29: Prefer to use English comments~

packages/builder/builder-rspack-provider/src/types/config/security.ts
9aoy: SRI is difficult to support，because `webpack-subresource-integrity` has a hard code dependency on `webpack-html-plugin`.

9aoy: But we can support `security.checkSyntax` & `security.checkPolyfill` next version ~

chenjiahan: get 👍🏻

zllkjc: So, if we can't support security.sri in the near future, we can move the type from builder-share to webpack-provider~

chenjiahan: +1

packages/runtime/plugin-runtime/src/ssr/serverRender/utils.ts
zllkjc: done

chenjiahan (approved):

PR #3735, opened 2023-05-23
feat(app-tools): SSR mode support async entry

2heal1:
## Summary

support `source.enableAsyncEntry` works on ssr mode

## Details

`source.enableAsyncEntry` will enable async entry , and it works smoothly on csr mode .

But ssr mode doesn't support it ,cause it need some change with entry code and server consumer .

## Checklist

### DIFF
--- a/.changeset/chilly-ducks-perform.md
+++ b/.changeset/chilly-ducks-perform.md
@@ -4,3 +4,4 @@
 ---
 
 chore: ssr mode support async entry
+chore: ssr 模式支持异步入口
--- a/.changeset/chilly-ducks-perform.md
+++ b/.changeset/chilly-ducks-perform.md
@@ -3,5 +3,5 @@
 '@modern-js/prod-server': patch
 ---
 
-chore: ssr mode support async entry
-chore: ssr 模式支持异步入口
+feat: ssr mode support async entry
+feat: ssr 模式支持异步入口
--- a/.changeset/chilly-ducks-perform.md
+++ b/.changeset/chilly-ducks-perform.md
@@ -1,0 +1,6 @@
+---
+'@modern-js/app-tools': patch
+'@modern-js/prod-server': patch
+---
+
+chore: ssr mode support async entry
--- a/packages/server/prod-server/src/libs/render/ssr.ts
+++ b/packages/server/prod-server/src/libs/render/ssr.ts
@@ -78,7 +78,7 @@ export const render = async (
   context.metrics = createMetrics(context, ctx.metrics);
 
   runner.extendSSRContext(context);
-  const bundleJSContent = await require(bundleJS);
+  const bundleJSContent = await Promise.resolve(require(bundleJS));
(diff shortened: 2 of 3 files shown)


.changeset/chilly-ducks-perform.md
```diff
@@ -0,0 +1,7 @@
+---
+'@modern-js/app-tools': patch
+'@modern-js/prod-server': patch
+---
+
+chore: ssr mode support async entry
```
chenjiahan, line 6: chore -> feat

packages/server/prod-server/src/libs/render/ssr.ts
```diff
@@ -78,7 +78,8 @@ export const render = async (
   context.metrics = createMetrics(context, ctx.metrics);
 
   runner.extendSSRContext(context);
-  const serverRender = require(bundleJS)[SERVER_RENDER_FUNCTION_NAME];
+  const bundleJSContent = await require(bundleJS);
```
chenjiahan, line 81: Is `await require` a valid usage? How about using `await import`?

packages/solutions/app-tools/src/analyze/generateCode.ts
```diff
@@ -285,9 +285,31 @@ export const generateCode = async (
...
+          entryName,
+          config.server.ssr,
+          config.server.ssrByEntries,
+          packageName,
+        );
+        if (ssr) {
+          rawAsyncEntryCode = `
+          export const serverRender = async (...args) => {
+            let exports = await ${rawAsyncEntryCode};
+            if(exports.default instanceof Promise){
```
chenjiahan, line 299:
```suggestion
            if (exports.default instanceof Promise){
```

.changeset/chilly-ducks-perform.md
2heal1: done

packages/solutions/app-tools/src/analyze/generateCode.ts
```diff
@@ -285,9 +285,31 @@ export const generateCode = async (
...
+        let rawAsyncEntryCode = `import('./${ENTRY_BOOTSTRAP_FILE_NAME}');`;
+        const ssr = getEntryOptions(
+          entryName,
+          config.server.ssr,
+          config.server.ssrByEntries,
+          packageName,
+        );
+        if (ssr) {
+          rawAsyncEntryCode = `
+          export const serverRender = async (...args) => {
```
chenjiahan, line 297: serverRender -> SERVER_RENDER_FUNCTION_NAME

packages/solutions/app-tools/src/analyze/generateCode.ts
```diff
@@ -285,9 +285,31 @@ export const generateCode = async (
...
+            let exports = await ${rawAsyncEntryCode};
```
chenjiahan, line 298: `exports` is a special object that is used in Node.js modules, maybe we can use other variable names.

PR #3760, opened 2023-05-25
fix(prod-server): the ssr render maybe get invalid url

GiveMe-A-Name:
## Summary

copilot:summary

## Details

copilot:walkthrough

## Checklist

### DIFF
--- a/packages/server/prod-server/src/libs/context/context.ts
+++ b/packages/server/prod-server/src/libs/context/context.ts
@@ -154,7 +154,10 @@ export class ModernServerContext implements ModernServerContextInterface {
     if (!host) {
       host = this.getReqHeader('Host');
     }
-    return (host as string).split(/\s*,\s*/, 1)[0] || '';
+    // the host = '',if we can't cat Host or X-Forwarded-Host header
+    // but the this.href would assignment `http[s]://${pathname}`
+    // the url is invalid, so we need assign host a no-empty value.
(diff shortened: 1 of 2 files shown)


packages/server/prod-server/src/libs/context/context.ts
```diff
@@ -154,7 +154,10 @@ export class ModernServerContext implements ModernServerContextInterface {
     if (!host) {
       host = this.getReqHeader('Host');
     }
-    return (host as string).split(/\s*,\s*/, 1)[0] || '';
+    // the host = '',if we can't cat Host or X-Forwarded-Host header
+    // but the this.href would assignment `http[s]://${pathname}`
+    // the url is invalid, so we need assign host a no-empty value.
+    return (host as string).split(/\s*,\s*/, 1)[0] || 'undefined';
```
GiveMe-A-Name, line 160: the no-empty value may can be `localhost:xxx`, I not sure, Maybe.

zllkjc: This looks like a break change when user code is `if(ctx.host) {}`, can it be handled where the `new URL()` is used

GiveMe-A-Name: It may be a fix in a different opinion.
If `new URL()` would occur `Invalid URL` before, it also means the `new URL()` can't used in code. So we can think `new URL()` no-exist in page codes.

zllkjc (approved):

PR #3761, opened 2023-05-25
fix(plugin-esbuild): set minify format to iife when target is web

chenjiahan:
## Summary

Keep the default format the same as `esbuild-loader` v3. We cannot bump esbuild-loader v3 now because it requires Node.js >= 16.

> The default is iife when esbuild is configured to support a low target, because esbuild injects helper functions at the top of the code. On the web, having functions declared at the top of a script can pollute the global scope. In some cases, this can lead to a variable collision error. By setting format: 'iife', esbuild wraps the helper functions in an [IIFE](https://developer.mozilla.org/en-US/docs/Glossary/IIFE) to prevent them from polluting the global.

copilot:summary

## Details

copilot:walkthrough

## Related Issue

https://github.com/esbuild-kit/esbuild-loader/commit/a9e8e7e1630ab11371f88f9349cad7ebbce57efc

## Checklist

### DIFF
--- a/.changeset/quick-hairs-doubt.md
+++ b/.changeset/quick-hairs-doubt.md
@@ -1,0 +1,7 @@
+---
+'@modern-js/builder-plugin-esbuild': patch
+---
+
+fix(plugin-esbuild): set minify format to iife when target is web
+
+fix(plugin-esbuild): 当 target 为 web 时，将压缩的 format 设置为 iife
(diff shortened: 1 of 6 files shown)


chenjiahan:
@Asuka109 The reason for the change is explained in the PR description
image

Asuka109:
okk, got it🤣

Asuka109 (approved):

PR #3991, opened 2023-06-16
feat: source build mode

targeral:
### <samp>🤖 Generated by Copilot at 9dc6956</samp>

This pull request introduces a new `plugin-source-build` package that enables source code build mode for monorepo projects. The package provides a builder plugin that modifies the webpack config to resolve modules based on the closest `tsconfig.json` file. The pull request also adds support for the plugin in the `app-tools` package and updates the `package.json` script and the `turbo` cache. The pull request also deletes an unused file and adds some ESLint and vitest configurations for the new package.

### <samp>🤖 Generated by Copilot at 9dc6956</samp>

*  Add a new builder plugin `plugin-source-build` to enable source code build mode for monorepo projects (F5-F16)

## Checklist


packages/toolkit/monorepo-utils/package.json
```diff
@@ -0,0 +1,59 @@
...
+  },
+  "scripts": {
+    "prepublishOnly": "only-allow-pnpm",
+    "new": "modern-lib new",
+    "dev": "modern-lib build --watch",
+    "build": "modern-lib build",
+    "test": "jest --passWithNoTests"
+  },
+  "dependencies": {
+    "@microsoft/rush-lib": "^5.99.0",
```
chenjiahan, line 43: `@microsoft/rush-lib` contains so many sub-dependencies, maybe we should move it to devDependencies.

packages/toolkit/monorepo-utils/package.json
```diff
@@ -0,0 +1,59 @@
...
+    "@modern-js/utils": "^2",
```
chenjiahan, line 44:
```suggestion
    "@modern-js/utils": "workspace:*",
```

packages/builder/builder-webpack-provider/src/config/validate/experiments.ts
```diff
@@ -2,4 +2,8 @@ import { sharedExperimentsConfigSchema, z } from '@modern-js/builder-shared';
 import type { ExperimentsConfig } from '../../types';
 
 export const experimentsConfigSchema: z.ZodType<ExperimentsConfig> =
-  sharedExperimentsConfigSchema;
+  sharedExperimentsConfigSchema
+    .extend({
+      sourceBuild: z.boolean(),
```
chenjiahan, line 7: If the `sourceBuild` config is built in, we should also make `builder-plugin-source-build` as a built-in plugin.

But `builder-plugin-source-build` now has some third-party dependencies, we do not want to increase the install size of builder, so we need to reduce the dependencies if we want to build in it.

targeral: I think some third-party dependencies should be required, such as @pnpm/filter-workspace-packages, @pnpm/logger. So we turn `builder-plugin-source-build` into an external plugin?

packages/toolkit/monorepo-utils/package.json
targeral: With these considerations in mind, perhaps we should use builder-plugin-build-source as an external plugin.

packages/toolkit/monorepo-utils/package.json
targeral: OK

packages/builder/builder-webpack-provider/src/config/validate/experiments.ts
targeral: I have removed some third-party dependencies from builder-plugin-source-build

packages/toolkit/monorepo-utils/package.json
targeral: `@microsoft/rush-lib` has been removed from the project.

packages/builder/builder-webpack-provider/src/config/validate/experiments.ts
chenjiahan: Yes, we can turn `builder-plugin-source-build` into an external plugin now.

chenjiahan (approved):

Issue #3889, opened 2023-06-08
[Bug]: plugin-testing not work


KyrieLii:
### Version

```sh
System:
    OS: macOS 12.5.1
    CPU: (10) arm64 Apple M1 Pro
    Memory: 74.67 MB / 32.00 GB
    Shell: 5.8.1 - /bin/zsh
  Browsers:
    Chrome: 114.0.5735.106
    Safari: 15.6.1
  npmPackages:
    @modern-js/app-tools: 2.22.0 => 2.22.0 
    @modern-js/eslint-config: 2.22.0 => 2.22.0 
    @modern-js/plugin-testing: 2.22.0 => 2.22.0 
    @modern-js/runtime: 2.22.0 => 2.22.0 
    @modern-js/tsconfig: 2.22.0 => 2.22.0
```

### Details

Follow the [Doc](https://modernjs.dev/en/guides/advanced-features/testing.html) , enable testing , and import tesing api:
```ts filename=index.test.tsx
import { render } from '@modern-js/runtime/testing';

describe('Default cases', () => {
  test('hello world test', () => {
    const s = 'hello Modern.js';
    render(<div>12</div>);
    expect(s).toBe('hello Modern.js');
  });
});
```

Run `pnpm test`: 

<img width="1245" alt="image" src="https://github.com/web-infra-dev/modern.js/assets/16858738/df241394-f78a-4c4d-a58e-455b3e8f512d">

### Reproduce link

https://modernjs.dev/en/guides/advanced-features/testing.html

### Reproduce Steps

See Details.

PR #4070, opened 2023-06-27
fix(testing): jest unexpected token error for @modern-js/runtime

targeral:
### <samp>🤖 Generated by Copilot at a165586</samp>

This pull request fixes a jest error for `@modern-js/runtime` by adding a `transformIgnorePatterns` option to the `@modern-js/plugin-testing` package. It also updates the changeset file for the patch release of the package.

### <samp>🤖 Generated by Copilot at a165586</samp>

*  Add a changeset file to document the patch updates for the `@modern-js/plugin-testing` package (.changeset/orange-horses-wonder.md)
*  Add a `transformIgnorePatterns` option to the `testingPlugin` function to exclude the `@modern-js/runtime` package from being transformed by jest ([link](https://github.com/web-infra-dev/modern.js/pull/4070/files?diff=unified&w=0#diff-03280beb0bbf5eaa5166abe24f86095e69cd4bbb15fbf4cdbd29f294b8a060aeR131-R133))

## Related Issue

#3889

## Checklist

### DIFF
--- a/packages/runtime/plugin-testing/src/cli/index.ts
+++ b/packages/runtime/plugin-testing/src/cli/index.ts
@@ -26,6 +26,26 @@ export const mergeUserJestConfig = (testUtils: TestConfigOperator) => {
   }
 };
 
+export const getJestTransformEsModulesRegStr = () => {
+  const esmModulesInPnpm = [
+    '@modern-js\\+runtime@',
+    '@modern-js\\+plugin@',
+    // @modern-js-reduck+store, @modern-js-reduck+effects and so on
+    '@modern-js-reduck',
+    '@babel\\+runtime@',
+  ];
+  // yarn or npm
+  const esmModules = [
+    '@modern-js/runtime',
+    '@modern-js/plugin',
+    '@modern-js-reduck',
+    '@babel/runtime',
+  ];
+  return `node_modules/(?!(\\.pnpm/(${esmModulesInPnpm.join(
+    '|',
+  )}))|(${esmModules.join('|')}))`;
+};
+
 export const testingPlugin = (): CliPlugin<{
   hooks: Hooks;
   userConfig: UserConfig;
@@ -128,6 +148,7 @@ export const testingPlugin = (): CliPlugin<{
               `<rootDir>/src/**/*.test.[jt]s?(x)`,
               `<rootDir>/tests/**/*.test.[jt]s?(x)`,
             ],
+            transformIgnorePatterns: [getJestTransformEsModulesRegStr()],
           });
 
           mergeUserJestConfig(utils);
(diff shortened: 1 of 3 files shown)


packages/runtime/plugin-testing/src/cli/index.ts
```diff
@@ -128,6 +128,9 @@ export const testingPlugin = (): CliPlugin<{
               `<rootDir>/src/**/*.test.[jt]s?(x)`,
               `<rootDir>/tests/**/*.test.[jt]s?(x)`,
             ],
+            transformIgnorePatterns: [
```
chenjiahan, line 131: Some users may not use pnpm, so we should also match paths of npm and yarn.

targeral: I'll try npm and yarn.

targeral: Done. You can check it.

packages/runtime/plugin-testing/src/cli/index.ts
```diff
@@ -26,6 +26,26 @@ export const mergeUserJestConfig = (testUtils: TestConfigOperator) => {
...
+    '@babel\\+runtime@',
+  ];
+  // yarn or npm
+  const esmModules = [
+    '@modern-js/runtime',
+    '@modern-js/plugin',
+    '@modern-js-reduck',
+    '@babel/runtime',
+  ];
+  return `node_modules/(?!(\\.pnpm/(${esmModulesInPnpm.join(
```
chenjiahan, line 44: Is `.pnpm` required? Maybe remove it can work well in pnpm projects

chenjiahan: Pnpm will create two layers of `node_modules`, such as: `node_modules/.pnpm/@modern-js+runtime@0.0.0/node_modules/@modern-js/runtime/package.json`

So remove the `\\.pnpm/` should work.

targeral: I removed `\\.pnpm/` and the test case didn't pass. Can you give me an example?

targeral: > It will not match `'node_modules/.pnpm/@modern-js+runtime@0.0.0'`.

I made a mistake here, when `\\.pnpm/` is removed, it matches to `'node_modules/.pnpm/@modern-js+runtime@0.0.0'`, but the result we want is a mismatch. Only if it doesn't match will jest process the package.

targeral: The rule here is to match all paths that do not contain `node_modules/.pnpm/@modern-js+runtime@`

chenjiahan (approved):

packages/runtime/plugin-testing/src/cli/index.ts
chenjiahan: ok, I get it👌

PR #4276, opened 2023-07-24
fix: upgrade tools build config

caohuilin:
### <samp>🤖 Generated by Copilot at 10c4360</samp>

Modified the build configuration of the `@modern-js/toolkit` package to optimize it for the modern.js framework. Disabled some features and set an alias for the `chalk` module in `modern.config.js`.

### <samp>🤖 Generated by Copilot at 10c4360</samp>

*  Customize build configuration for `@modern-js/toolkit` package ([link](https://github.com/web-infra-dev/modern.js/pull/4276/files?diff=unified&w=0#diff-8f17c56df193f8c3e4fb0d73a55f9cd3bbc8200feef028bbbd4b0bf70aa76049L1-R9))

## Checklist


KyrieLii (dismissed):

packages/toolkit/upgrade/modern.config.js
```diff
@@ -1,5 +1,10 @@
-const { tscLikeBuildConfig } = require('@scripts/build');
-
 module.exports = {
-  buildConfig: tscLikeBuildConfig,
+  buildConfig: {
+    autoExternal: false,
```
chenjiahan, line 3: If the build config is updated, maybe the dependencies should be updated too

packages/toolkit/upgrade/modern.config.js
```diff
@@ -1,5 +1,10 @@
...
+    alias: {
+      chalk: '@modern-js/utils/chalk',
```
chenjiahan, line 5: So why do we need to bundle this package? It will cause duplicate code.

For example, `@modern-js/codesmith` will be installed in the user's project, but it will also be bundled with `@modern-js/upgrade`. There will be two copies of the `@modern-js/codesmith` code.

caohuilin: I get you mean. `@modern-js/upgrade` publish file now is only `dist/index.js` to run with `npx @modern-js/upgrade`. We need to support two methods.

chenjiahan: So why does `npx @modern-js/upgrade` need to run with only `dist/index.js`?

chenjiahan (approved):

PR #4391, opened 2023-08-07
fix: typing error because the `app` is required (but actually not)

Asuka109:
### <samp>🤖 Generated by Copilot at cbd6dd7</samp>

Refactor `Server` initialization code in `index.ts` to use `Object.assign` instead of destructuring. This enables the `runner` to return additional properties for the `Server` instance.

### <samp>🤖 Generated by Copilot at cbd6dd7</samp>

*  Refactor server initialization to use `Object.assign` instead of destructuring assignment ([link](https://github.com/web-infra-dev/modern.js/pull/4391/files?diff=unified&w=0#diff-5cbeedfd701ea461dc9742e0bff36e00e9dd1faae4fb4dc686213499a62cb542L109-R123))

## Checklist

### DIFF
--- a/packages/server/prod-server/src/server/index.ts
+++ b/packages/server/prod-server/src/server/index.ts
@@ -106,21 +106,21 @@ export class Server {
     }
 
     // runner can only be used after server init
-    Object.assign(
-      this,
-      await this.runner.beforeServerInit({
+    {
+      const result = await this.runner.beforeServerInit({
         app: this.app,
         server: this.server,
-      }),
-    );
+      });
+      ({ app: this.app = this.app, server: this.server } = result);
+    }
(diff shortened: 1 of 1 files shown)


packages/server/prod-server/src/server/index.ts
```diff
@@ -106,18 +106,21 @@ export class Server {
     }
 
     // runner can only be used after server init
-    ({ app: this.app, server: this.server } =
+    Object.assign(
```
chenjiahan, line 109: Explicit assignment would be better

Asuka109: Got it.

chenjiahan (approved):

PR #4531, opened 2023-08-28
feat(builder): support enableCssModuleTSDeclaration in rspack mode

9aoy:
## Summary

handle css modules like [css-loader](https://github.com/webpack-contrib/css-loader/blob/4673caa4aa68d5fb1127c172b4afd081bd56eb73/src/utils.js#L759), but only get cssModuleKeys and generator cssModuleTSDeclaration, not modify the source content.

use these packages to deal with css modules and extract [icss](https://github.com/css-modules/icss) exports:
- postcss
- postcss-modules-extract-imports
- postcss-modules-local-by-default
- postcss-modules-scope
- postcss-modules-values
- icss-utils

### <samp>🤖 Generated by Copilot at 9fdfc3e</samp>

No summary available (Limit exceeded: required to process 66178 tokens, but only 50000 are allowed per call)

### <samp>🤖 Generated by Copilot at 9fdfc3e</samp>

No walkthrough available (Limit exceeded: required to process 66178 tokens, but only 50000 are allowed per call)

## Checklist


packages/builder/builder-shared/src/css.ts
```diff
@@ -195,6 +197,9 @@ export const normalizeCssLoaderOptions = (
   return options;
 };
 
+export const getCssModuleLocalsConvention = (config: SharedNormalizedConfig) =>
+  config.output.cssModules?.exportLocalsConvention || 'camelCase';
```
chenjiahan, line 201: Should we move this default config to the default builder config?

packages/builder/builder-rspack-provider/src/rspackLoader/css-modules-typescript-pre-loader/index.ts
```diff
@@ -0,0 +1,102 @@
...
+  const { cssModuleKeys } = await processCss(content, {
+    exportLocalsConvention: modules.exportLocalsConvention,
+  });
+
+  // @ts-expect-error
+  this.cssModuleKeys = cssModuleKeys;
+
+  // rspack loader type is not exactly the same as webpack loader type
+  // @ts-expect-error
+  return cssModulesTypescriptLoader.call(this, content, ...input);
```
h-a-n-a, line 101: Isn't the `content` be the result of `css-loader`? Isn't it the necessity of this loader?

chenjiahan (approved):

packages/builder/builder-rspack-provider/src/rspackLoader/css-modules-typescript-pre-loader/index.ts
9aoy: we not use css-loader in rspack mode, the content is the result of other loaders(less/sass/postcss-loader).

we need to get css modules keys, but we can't get the result of experiments.css, so we need to fake a css-modules result

h-a-n-a (approved):

PR #4795, opened 2023-10-16
chore(workflow): integrate nx as monorepo solution

bvanjoi:


nx.json
```diff
@@ -0,0 +1,26 @@
...
+        "{projectRoot}/tsconfig.json",
+        "{projectRoot}/package.json",
+        "{projectRoot}/modern.config.*",
+        "{projectRoot}/scripts/**/*"
+      ]
+    }
+  },
+  "tasksRunnerOptions": {
+    "default": {
+      "runner": "nx-cloud",
```
chenjiahan, line 17: Will we pay for nx-cloud?

bvanjoi:
Failed to upload coverage report, is it common?

image

nx.json
bvanjoi: I aim to utilize nx cloud primarily to determine the causes of cache misses, rather than for remote caching. Consequently, I think it may not necessitate any payment.

package.json
```diff
@@ -77,13 +78,14 @@
...
     "@vitest/coverage-v8": "^0.33.0",
+    "@vitest/ui": "^0.33.0",
     "esbuild": "0.17.19",
     "eslint": "^8.28.0",
-    "lint-staged": "~13.1.0",
     "husky": "^8.0.0",
-    "turbo": "1.10.2",
+    "lint-staged": "~13.1.0",
+    "nx": "16.10.0",
+    "nx-cloud": "^16.5.2",
```
chenjiahan, line 88: if we do not use `nx-cloud`, can we remove it from dependencies? this can reduce the dependencies number.

bvanjoi: It has indeed been used.

chenjiahan (approved):

Issue #4780, opened 2023-10-12
Does the coverage test truly provide help?


bvanjoi:
### Version

```sh
nothing
```

### Details

After reviewing this [comment](https://github.com/web-infra-dev/modern.js/pull/4775#issuecomment-1757374793), I am skeptical about the real benefit of this coverage for our testing.

I also noticed several potential issues:
- We are using ﻿`vitest` and `﻿jest` concurrently, and `﻿vitest` by default employs ﻿`c8` to generate coverage, which is not as precise as the ﻿babel coverage used by ﻿`jest`.
- Can the coverage report genuinely assist us in identifying bugs we've overlooked? I believe for most scenarios, simply adding relevant tests should be sufficient.

### Reproduce link

nothing

### Reproduce Steps

nothing

chenjiahan:
I have the same opinion as you, the coverage is inaccurate and probably useless now.

Our tests include Jest UT, Vitest UT, Jest E2E and Playwright E2E, each using separate coverage reports. In my daily development, the coverage reports are often of no use.

If disabling coverage can improve the test performance, we can consider changing it.

chenjiahan:
Also, based on my experience in other projects, Vitest should use Istanbul to provide more accurate coverage.

PR #4811, opened 2023-10-18
chore: disable coverage in ci

bvanjoi:
close https://github.com/web-infra-dev/modern.js/issues/4780

### DIFF
--- a/packages/builder/builder-cli/package.json
+++ b/packages/builder/builder-cli/package.json
@@ -38,7 +38,7 @@
     "build": "modern-lib build",
     "dev": "modern-lib build --watch",
     "test": "vitest run",
-    "test:watch": "vitest dev --no-coverage"
+    "test:watch": "vitest dev"
   },
   "dependencies": {
     "@modern-js/builder": "workspace:*",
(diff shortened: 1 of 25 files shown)


bvanjoi:
It can reduce the time cost.

before: https://github.com/web-infra-dev/modern.js/actions/runs/6557895579/job/17810369144?pr=4809

image

now: 

image

packages/builder/builder-cli/package.json
```diff
@@ -37,7 +37,7 @@
     "new": "modern new",
     "build": "modern-lib build",
     "dev": "modern-lib build --watch",
-    "test": "vitest run",
+    "test": "vitest run --no-coverage",
```
chenjiahan, line 40: Why not just remove https://github.com/web-infra-dev/modern.js/blob/main/scripts/vitest-config/src/index.ts#L8-L10

bvanjoi: I did not notice that.

chenjiahan: Haha it's a bit hidden

chenjiahan (approved):

PR #4862, opened 2023-10-25
refactor: simpify `false.test.ts`

bvanjoi:
This PR introduces two changes:
- Use self-hosted runner for Linux integration tests
- Simply the `﻿resolve/false.test.ts` to see if they improve in Windows.

---

And there contains two case in the most failed ci recently:

- `module/build/resolve/false.test.ts` in windows.
- timeout in linux

(I think there's a 70% chance that the issue is **not** caused by nx.)

Regardless, let's proceed in trying to solve this problem

### DIFF
--- a/.github/workflows/integration-test-Linux.yml
+++ b/.github/workflows/integration-test-Linux.yml
@@ -13,17 +13,22 @@ on:
 jobs:
   integration-test-linux:
     # The type of runner that the job will run on
-    runs-on: ubuntu-latest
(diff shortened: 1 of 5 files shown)


chenjiahan (approved):

10Derozan (approved):

10Derozan:
`module/build/resolve/false.test.ts` failed in windows because #4856 , i've fixed it.

PR #5004, opened 2023-11-27
fix(builder): preload resource can't be used if crossOrigin attr not match

9aoy:
## Summary

fix preload resource can't be used when crossOrigin attr does not match.
image

## Checklist


9aoy:
!canary

chenjiahan (approved):

PR #5021, opened 2023-11-30
fix: bff-core should be a dependency of testing plugin

yimingjfe:
## Checklist


clChenLiang (approved):
LGTM

PR #5032, opened 2023-12-04
feat: ssr add more context

GiveMe-A-Name:
## Summary

In some situations, ssr need some information from req, res.
So we need support server.ssr.unsafeContext.
we support more information put window._SSR_DATA

## Checklist


packages/server/core/src/types/config/server.ts
```diff
@@ -41,6 +41,9 @@ export type SSR =
       preload?: boolean | SSRPreload;
       inlineScript?: boolean;
       disablePrerender?: boolean;
+      unsafeContext?: {
+        headers?: string[];
```
zllkjc, line 45: Are there any other fields here?

GiveMe-A-Name: Haven not.
unsafeContext is an Object because of the request have many attributes. 
But I think this fields can be `unsafeHeader`... ?
What about your proposal?

GiveMe-A-Name:
!canary

zllkjc (approved):

PR #5061, opened 2023-12-11
feat(server): add ModernRsbuildServer

9aoy:
## Summary

The old modern server should be replaced by ModernRsbuildServer when rsbuild is fully integrated into modern.js.

## Checklist


packages/server/server/src/server/devServer.ts
```diff
@@ -84,48 +115,65 @@ export class ModernDevServer extends ModernServer {
...
-    const { befores, afters } = this.applySetupMiddlewares();
+    this.rsbuild.onDevCompileDone(({ stats }) => {
+      // Reset only when client compile done
+      if (stats.toJson({ all: false }).name !== 'server') {
+        this.onRepack({ routes: this.getRoutes() });
+      }
+    });
 
     // before dev handler
     const beforeHandlers = await this.setupBeforeDevMiddleware();
```
zllkjc, line 164: It seems that the order here is inconsistent.
Origin: before - > hookBefore - > unshift - > push - > hookAfter - > after
Now: hookBefore - > before - > unshift - > push - > after - > hookAfter

However, these two hooks are not used. we can remove them, and developer won't need to use them later.

packages/server/server/src/server/devServerOld.ts
```diff
@@ -0,0 +1,462 @@
+import { createServer, IncomingMessage, Server, ServerResponse } from 'http';
```
zllkjc, line 1: remove this file ? and also remove the dependencies like "http-compression"

packages/server/server/src/server/devServer.ts
9aoy: emm... I will remove these two hooks  in next PR

packages/server/server/src/server/devServerOld.ts
9aoy: The old modern server and deps should be replaced by ModernRsbuildServer after rsbuild is fully integrated into modern.js. Now, we still need it.

packages/server/server/src/server/devServer.ts
9aoy: updated!

zllkjc (approved):

PR #5065, opened 2023-12-11
fix: temp routes file should use alias intead of absolute path

yimingjfe:
## Checklist


packages/solutions/app-tools/src/analyze/nestedRoutes.ts
```diff
@@ -194,24 +194,24 @@ export const walk = async (
 
     if (itemWithoutExt === NESTED_ROUTE.LAYOUT_LOADER_FILE) {
       if (!route.loader) {
-        route.loader = itemPath;
+        route.loader = replaceWithAlias(alias.basename, itemPath, alias.name);
```
clChenLiang, line 197: why not up the code `replaceWithAlias(alias.basename, itemPath, alias.name)` to upper scope.
It seems use the same params to get same result

clChenLiang (approved):

packages/solutions/app-tools/src/analyze/nestedRoutes.ts
yimingjfe: done~

clChenLiang (approved):

PR #5087, opened 2023-12-13
fix(plugin-swc): should not override extensions when using function config

JSerFeng:
## Summary

fix function config overrides user extension config
fix function config return undefined causes final config to be undefined

## Checklist


chenjiahan (approved):

chenjiahan:
Can you also submit a PR to the Rsbuild repo? https://github.com/web-infra-dev/rsbuild/tree/main/packages/compat/plugin-swc

We will merge these two copies of the code soon.

PR #5119, opened 2023-12-19
fix(module-tools): clear .tsbuildinfo before build instead of tsc --clean

10Derozan:
## Summary
In #5064 , we execa `tsc --build --clean`, the reason is that when we clean the output, we doesn't clean the .tsbuildinfo which generated by incremental build. 
But `tsc --build --clean` will bring two problems: 
- first, tsc --clean will remove the outputs by esbuild, tsc --clean only respect the project tsconfig.json, so this need that we should set `emitDeclarationOnly` in our tsconfig.
- second, we don't respect the incremental build. we delete the outputs for all projects.

Now we will respect the incremental, in the root project, we will clear the .tsbuildinfo, unless you close the clean func.
For the dependencies, we do nothing.

## Related Links

#5064 

## Checklist

### DIFF
--- a/.changeset/silly-timers-knock.md
+++ b/.changeset/silly-timers-knock.md
@@ -1,0 +1,6 @@
+---
+'@modern-js/module-tools': patch
+---
+
+fix(module-tools): clear .tsbuildinfo before build instead of tsc --clean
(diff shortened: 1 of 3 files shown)


chenjiahan (approved):

chenjiahan:
👍🏻

PR #5168, opened 2023-12-28
feat(app-tools): use uni-builder instead of modern builder

9aoy:
## Summary

use uni-builder(rsbuild) instead of modern.js builder, modern.js builder is no longer maintained and will be removed later
   - use uni-builder instead of modern.js builder in app-tools and related pkgs
   - unify AppToolsUserConfig

The `@modern-js/uni-builder` package provides the same capability as `@modern-js/builder` + `@modern-js/builder-webpack-provider` + `@modern-js/rspack-provider`.

## Checklist


packages/solutions/app-tools/src/config/default.ts
```diff
@@ -1,23 +1,26 @@
...
   const dev: AppUserConfig['dev'] = {
-    ...defaultBuilderConfig.dev,
     // `dev.port` should not have a default value
     // because we will use `server.port` by default
     port: undefined,
   };
 
   const output: AppUserConfig['output'] = {
-    ...defaultBuilderConfig.output,
+    distPath: {
```
chenjiahan, line 16: why do we need to re-define default config here

9aoy: api.useResolvedConfigContext requires it. api.useResolvedConfigContext needs to be set & used regardless of whether builder config is init / modify or not. 
I don't think this is a good design, but it's a bit hard to change...

package.json
```diff
@@ -27,6 +27,7 @@
     "lint:package-json": "cd ./scripts/lint-package-json && pnpm start",
     "prepare-build": "cross-env NX_DAEMON=false nx run-many -t build -p @modern-js/* --maxParallel=100",
     "prepare": "npm run prepare-build && husky install",
+    "build:required": "cross-env NX_DAEMON=false nx run-many -t build -p @modern-js/uni-builder",
```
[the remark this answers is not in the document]
9aoy: @modern-js/uni-builder's moduleResolution is different from the moduleResolution configuration of other packages, and there will currently be some errors.

aadbd8e5-11dd-4ddc-80a4-f14cb35b08f6

9aoy:
!bench

9aoy:
!bench framework

9aoy:
!bench-framework

Timeless0911:
!bench-framework

9aoy:
!bench-framework

9aoy:
!bench-framework

chenjiahan (approved):

PR #5174, opened 2023-12-29
fix(uni-builder): compat builder plugin api & type

9aoy:
## Summary

Builder plugin can be used in uni-builder if  the builder plugin does not depend on some incompatible config, such as output.disableCssExtract.

## Checklist

### DIFF
--- a/packages/builder/uni-builder/package.json
+++ b/packages/builder/uni-builder/package.json
@@ -73,6 +73,7 @@
     "webpack-subresource-integrity": "5.1.0"
   },
   "devDependencies": {
+    "@modern-js/builder-plugin-node-polyfill": "workspace:*",
     "@rsbuild/plugin-swc": "0.2.13",
     "@scripts/build": "workspace:*",
     "@scripts/vitest-config": "workspace:*",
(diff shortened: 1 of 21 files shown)


packages/builder/uni-builder/src/shared/compatLegacyPlugin.ts
```diff
@@ -0,0 +1,72 @@
...
+  UniBuilderPlugin,
+  UniBuilderPluginAPI,
+  UniBuilderContext,
+} from '../types';
+import type { RsbuildPlugin } from '@rsbuild/core';
+import { logger } from '@rsbuild/shared';
+import { join } from 'path';
+
+function addDeprecatedWarning(name: string, newName?: string) {
+  logger.warn(
```
chenjiahan, line 11: can we print the builder plugin name here

packages/builder/uni-builder/package.json
```diff
@@ -73,6 +73,7 @@
     "webpack-subresource-integrity": "5.1.0"
   },
   "devDependencies": {
+    "@modern-js/builder-plugin-node-polyfill": "workspace:*",
```
chenjiahan, line 76: this package will be removed soon, so we should not depend on it

9aoy: This will be removed along with other builder dependencies in modern.js repo.

chenjiahan (approved):

PR #5187, opened 2024-01-04
fix(runtime-utils): devServer served html content didn't change after repacking.

xc2:
## Summary

Fix problem: devServer served html content didn't change after repacking.

Affected versions: 2.43 to 2.45

Use `Container#delete` instead of `Storage#delete` to delete cached content as `Storage#keys` iters keys of container which is with namespace prefix.

## Checklist

- [x] I have updated the documentation: no need


GiveMe-A-Name:
Great, thank you for finding my mistake.

GiveMe-A-Name (approved):

PR #5226, opened 2024-01-15
perf(module-tools): skip build js and d.ts when input is empty

10Derozan:
## Summary
Avoid rollup error when input is empty.

## Checklist

### DIFF
--- a/packages/solutions/module-tools/src/builder/build.ts
+++ b/packages/solutions/module-tools/src/builder/build.ts
@@ -34,6 +34,11 @@ export const runBuildTask = async (
   const { watch } = buildCmdOptions;
   const existTsconfig = await fs.pathExists(buildConfig.tsconfig);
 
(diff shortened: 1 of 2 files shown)


packages/solutions/module-tools/src/builder/build.ts
```diff
@@ -34,6 +34,11 @@ export const runBuildTask = async (
   const { watch } = buildCmdOptions;
   const existTsconfig = await fs.pathExists(buildConfig.tsconfig);
 
+  if (Object.keys(buildConfig.input).length === 0) {
+    logger.info('If input is empty, js and dts task will be skipped');
```
chenjiahan, line 38:
```suggestion
    logger.info('The input config is empty, as a result, the JS compilation and dts generation tasks will be skipped. Please provide input to run these tasks.');
```

chenjiahan (approved):

PR #5230, opened 2024-01-16
perf(module-tools): log error detail which may throw by own plugin and complete error stack

10Derozan:
### before
image

### after
image

And then we found that our error stack is complete. The ModuleBuildError has extended the esbuild error stack.

## Checklist

### DIFF
--- a/packages/solutions/module-tools/src/builder/esbuild/index.ts
+++ b/packages/solutions/module-tools/src/builder/esbuild/index.ts
@@ -9,6 +9,7 @@ import {
   ImportKind,
   formatMessages,
   Format,
+  BuildFailure,
 } from 'esbuild';
 import * as tapable from 'tapable';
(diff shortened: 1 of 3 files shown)


packages/solutions/module-tools/src/builder/esbuild/index.ts
```diff
@@ -292,6 +293,14 @@ export class EsbuildCompiler implements ICompiler {
         });
       }
     } catch (error: any) {
+      if ('errors' in error) {
```
targeral, line 296: Suggest:
``` js
if (error instanceof BuildFailure && 'errors' in error) {
  error?.errors?.forEach(err => {
```

targeral (approved):

packages/solutions/module-tools/src/builder/esbuild/index.ts
10Derozan: maybe only need `instanceof BuildFailure`

PR #5250, opened 2024-01-18
fix(uni-builder): should print https url when enable dev.https

9aoy:
## Summary

image

## Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -136,7 +136,7 @@
       "source-map-url": "0.4.1",
       "urix": "0.1.0",
       "resolve-url": "0.2.1",
-      "devcert": "1.0.0",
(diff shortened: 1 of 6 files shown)


package.json
```diff
@@ -136,7 +136,7 @@
       "source-map-url": "0.4.1",
       "urix": "0.1.0",
       "resolve-url": "0.2.1",
-      "devcert": "1.0.0",
+      "devcert": "1.2.2",
```
chenjiahan, line 139: is the latest version deprecated?

9aoy: removed

chenjiahan (approved):

PR #5298, opened 2024-01-24
docs: update builderPlugin introduction

9aoy:
## Summary

Modern.js app-tools's builder has been upgraded to rsbuild, so builderPlugin configuration is used to configure rsbuildPlugin.

## Checklist


packages/document/main-doc/docs/en/configure/app/builder-plugins/list.mdx
```diff
@@ -0,0 +1,21 @@
+---
+sidebar_label: Plugin List
```
chenjiahan, line 2: the configure folder should only contain config document

chenjiahan (approved):

PR #5323, opened 2024-01-29
feat(devtools): use loader to redirect to welcome page

Asuka109:
## Checklist

### DIFF
--- a/packages/devtools/client/src/entries/client/routes/headers/page.data.ts
+++ b/packages/devtools/client/src/entries/client/routes/headers/page.data.ts
@@ -1,0 +1,7 @@
+import { redirect } from '@modern-js/runtime/router';
+import { $state } from './state';
+
+export const loader = async () => {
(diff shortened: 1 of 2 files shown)


packages/devtools/client/src/entries/client/routes/headers/page.data.ts
```diff
@@ -0,0 +1,7 @@
+import { redirect } from '@modern-js/runtime/router';
+import { $state } from './state';
+
+export const loader = async () => {
+  const service = await $state.service;
+  return redirect(service.rules ? './editor' : './welcome');
```
chenjiahan, line 6: should we redirect in a loader?

chenjiahan: https://reactrouter.com/en/main/fetch/redirect

chenjiahan (approved):

PR #5329, opened 2024-01-30
fix: type errors

GiveMe-A-Name:
## Checklist

### DIFF
--- a/packages/builder/uni-builder/package.json
+++ b/packages/builder/uni-builder/package.json
@@ -76,7 +76,6 @@
   },
   "devDependencies": {
     "@modern-js/builder-plugin-node-polyfill": "workspace:*",
-    "@modern-js/server-core": "workspace:*",
     "@rsbuild/plugin-swc": "0.3.7",
     "@scripts/build": "workspace:*",
     "@scripts/vitest-config": "workspace:*",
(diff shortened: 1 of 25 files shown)


packages/builder/uni-builder/package.json
```diff
@@ -76,6 +76,7 @@
   },
   "devDependencies": {
     "@modern-js/builder-plugin-node-polyfill": "workspace:*",
+    "@modern-js/server-core": "workspace:*",
```
yimingjfe, line 79: Why was this dependency included?

packages/server/core/package.json
```diff
@@ -31,7 +31,7 @@
       "default": "./dist/cjs/index.js"
     },
     "./base": {
-      "types": "./base/index.ts",
+      "types": "./dist/types/base/index.d.ts",
```
yimingjfe, line 34: It should not be modified here，please see publishConfig

GiveMe-A-Name: Is just for source-code compiler?

packages/builder/uni-builder/package.json
GiveMe-A-Name: uni-builder need ServerBaseOptions type from modern-js/server-core

packages/server/core/package.json
GiveMe-A-Name: uni-builder use tsc to compiler, it seems not support source-code compile.

yimingjfe (approved):

PR #5335, opened 2024-01-30
feat(devtools): preparation for integration of rsdoctor

Asuka109:
## Summary

- Update test cases of devtools.
- Add rsdoctor as optional peer dependencies.
- New entries runtime & node of devtools kit for different env.
- Use flatted to serialize data to avoid circular reference.

## Checklist


packages/devtools/kit/src/rsdoctor.ts
```diff
@@ -0,0 +1,37 @@
...
+  const { Url } = await import('@rsdoctor/utils/common');
+  if (Url.isUrl(url)) {
+    const resp = await fetch(url);
+    return resp.text();
+  }
+  // json string
+  return url;
+}
+
+export async function parseManifest(
```
chenjiahan, line 13: Where do we use this method?

Asuka109: It is used for parsing manifest stats of rsdoctor.
Pasted from [rsdoctor](https://github.com/web-infra-dev/rsdoctor/blob/da541bee23bf339132a4ce3c67a92ffbd1737986/packages/components/src/utils/request.ts#L43) and will be used in the next PR.
For now it just a placeholder.

chenjiahan: 1. Maybe we should discuss with Rsdoctor to get a public API for this.
2. So why this placeholder is added to this PR not next PR.

Asuka109: Yep, I remove it from this PR.

chenjiahan (approved):

PR #5339, opened 2024-01-31
fix: bind bff handlers

yimingjfe:
## Checklist

### DIFF
--- a/packages/server/core/src/base/middlewares/customServer.ts
+++ b/packages/server/core/src/base/middlewares/customServer.ts
@@ -90,12 +90,18 @@ export class CustomServer {
         }
       }
 
+      // TODO: reduce the number of calls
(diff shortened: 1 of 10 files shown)


packages/server/core/src/base/middlewares/customServer.ts
```diff
@@ -90,13 +90,13 @@ export class CustomServer {
         }
       }
 
+      await next();
```
GiveMe-A-Name, line 93: Why call next earlier than c.finalized `return`

yimingjfe: In the next, some middlewares may also want to end the response, such as the ssr handler for redirect.

GiveMe-A-Name (approved):

PR #5364, opened 2024-02-02
feat(devtools): use subset of Noto Sans for the logo and headings

Asuka109:
## Summary

image

Use subset of Noto Sans for the logo and headings instead of SVG images.

Generate the subset through https://font-subset.disidu.com/ to reduce font file size (2.6 MB -> 70 KB).

## Checklist


packages/devtools/client/src/styles/fonts.scss
```diff
@@ -0,0 +1,11 @@
+/*
+ * This subset is modified from the Noto Sans font,
+ * which is distributed under the Open Font License.
+ * You can learn more about it at the following link.
+ * https://openfontlicense.org/
+ * https://fonts.google.com/noto/specimen/Noto+Sans
+ */
+@font-face {
+  font-family: 'NotoSans';
+  src: url('@/assets/NotoSans-latin-subset.ttf');
```
chenjiahan, line 10: woff2 is smaller

chenjiahan: can we use woff2

Asuka109: Yep, I converted it to woff2 format, resulting in a significantly smaller size (70KB -> 38 KB).

chenjiahan (approved):

packages/devtools/client/src/styles/fonts.scss
chenjiahan: 👍

PR #5371, opened 2024-02-06
feat(storybook): use uni-builder instead of modern.js builder as build tool

9aoy:
## Summary

modern.js builder is no longer maintained, should use uni-builder or rsbuild instead.

## Checklist


packages/storybook/builder/src/core.ts
```diff
@@ -27,27 +28,24 @@ export async function getCompiler(
...
-    }
-  }
-
-  const builder = await createBuilder(provider, {
+  const builder = await createUniBuilder({
+    bundlerType,
     cwd,
     target: 'web',
-    framework: 'modern.js storybook',
+    frameworkConfigPath: res?.path ? res?.path : undefined,
```
chenjiahan, line 44:
```suggestion
    frameworkConfigPath: res?.path ? res.path : undefined,
```

packages/storybook/builder/package.json
```diff
@@ -56,23 +56,24 @@
   },
   "license": "MIT",
   "dependencies": {
-    "@modern-js/builder": "workspace:*",
-    "@modern-js/builder-shared": "workspace:*",
+    "@rsbuild/core": "0.4.3",
```
chenjiahan, line 59: is @rsbuild/core used?

packages/storybook/builder/src/types.ts
```diff
@@ -1,21 +1,18 @@
-import type { BuilderConfig as WebpackBuilderConfig } from '@modern-js/builder-webpack-provider';
-import type { BuilderConfig as RspackBuilderConfig } from '@modern-js/builder-rspack-provider';
-import { BuilderPlugin } from '@modern-js/builder-shared';
+import type { RsbuildPlugin } from '@rsbuild/shared';
```
chenjiahan, line 1: It is recommended to import types / methods from `@rsbuild/core`, as `@rsbuild/shared` is a private package of Rsbuild and may change in non-major versions.

packages/storybook/builder/package.json
9aoy: removed

chenjiahan (approved):

PR #5396, opened 2024-02-19
feat: add getRenderHandler & optimize types

yimingjfe:
## Checklist


packages/server/core/src/base/adapters/node/bff.ts
```diff
@@ -0,0 +1,46 @@
...
+  const { getMiddlewares, ...collector } = createMiddlewareCollecter();
+  const { runner } = server;
+  await runner.gather(collector);
+
+  const webOnly = await isWebOnly();
+  let handler: ServerNodeMiddleware;
+  const renderHandler = getRenderHandler(options);
+  if (webOnly) {
+    handler = async (c, next) => {
+      c.body('');
```
GiveMe-A-Name, line 30: Should we give a 404 response if `webOnly`?

yimingjfe: No, 404 will cause an error, which is not what users expect.

GiveMe-A-Name: Got it

PR #5420, opened 2024-02-21
feat(solutions): separate generator from solutions

### LINKED ISSUE #5359: [Enhancement]: Remove new-action and upgrade generator from solutions.

Unlike other commands(eg. dev build inspect .. ), these two commands do not depend on build capabilities or even whether dependencies are installed. But it brings the overhead of dependent installation.

I think they can be run independently to reduce unnecessary dependencies on initialization.

Further, they can not be 'generators', but just simple cli tools to become more lightweight.

New and upgrade commands work fine but are not installed on initialization.

KyrieLii:
## Summary
close #5359 

### 1. Background

This PR made some optimizations by separating the 'new' and 'upgrade' commands from solutions (such as app-tools).
There are several benefits to doing this:
- Reduce the direct installation dependency count of solutions to decrease installation time.
- Allowing the 'new' and 'upgrade' commands to run independently for easier optimization in the future.

####  2.1  Refactor @modern-js/new-action and @modern-js/upgrade to make them run independently.
Now you can use it by `npx`: 
```shell
npx @modern-js/new-action // same as npm run new  (modern new)

npx @modern-js/upgrade // same as npm run upgrade (modern upgrade)
```

After refactoring, the original invocation method remains unaffected.

####  2.2 Remove the above two dependencies from solutions and execute the CLI using the npx method.

To maintain version consistency, a new environment variable MODREN_JS_VERSION has been added during execution.

### 3. Drawback

The first time executing the command will take longer, as the time cost has shifted to the npx execution.
Whenever the version of solutions changes, it can be considered as the first time executing.

Next, We will optimize the execution speed of the cli based on this foundation.

## Checklist


KyrieLii:
!canary

KyrieLii:
!canary

KyrieLii:
!canary

packages/generator/new-action/src/run.ts
```diff
@@ -0,0 +1,47 @@
...
+import { getSolutionByDependance } from './utils';
+
+const main = async () => {
+  const program = new Command();
+  program
+    .option('--solution <solution>', 'solution', '')
+    .option('--config <config>', 'config', '{}')
+    .option('--root-path <rootPath>', 'project root path', '')
+    .action(async params => {
+      if (!params.solution) {
```
zllkjc, line 18: It looks like a breaking change. Will the `new` command for `module-tools` templates append `--solution` ?

zllkjc: I see..no problem

zllkjc (approved):

PR #5430, opened 2024-02-22
chore(app-tools): use builder.startDevServer instead of ModernOldServer when call api-only

9aoy:
## Summary

use builder.startDevServer instead of ModernOldServer when call with `--api-only`, ModernOldServer will be removed soon

## Checklist


9aoy:
Duplicate with server refactor PR, CC @yimingjfe

PR #5438, opened 2024-02-23
feat(core): add global cli mode

targeral:
## Checklist


.changeset/green-buckets-check.md
chenjiahan, line 1: Can you provide some context for this change in the PR summary?

PR #5491, opened 2024-03-05
chore(generator): remove legacy storybook generator

9aoy:
## Summary

`Visual Testing (Storybook)` is a legacy storybook generator, we should remove it, and only keep `StorybookV7` generator.
image

## Checklist


packages/generator/generator-common/src/locale/en.ts
```diff
@@ -27,10 +27,8 @@ export const EN_LOCALE = {
       i18n: 'Enable Internationalization (i18n)',
       test: 'Enable Unit Test / Integration Test',
       e2e_test: 'Enable E2E Test',
-      storybook: 'Enable Storybook',
       storybookV7: 'Enable Storybook V7',
```
chenjiahan, line 30: Should we rename the 'Enable Storybook V7' to 'Enable Storybook'?

9aoy: In legacy generator 'Enable Storybook' means enable Storybook v6, I'm afraid if people read the new document and use legacy generator, they enable Storybook v6 by mistake.

chenjiahan: get

chenjiahan (approved):

PR #5493, opened 2024-03-05
fix(uni-builder): should not apply babel-loader when use rspack build

9aoy:
## Summary

cacheIdentifier is an addition config and it should ignore when we check isBabelConfigModified.

## Checklist


9aoy:
!bench-framework

chenjiahan (approved):

PR #5537, opened 2024-03-15
docs: add type checker range FAQ

chenjiahan:
## Summary

Add type checker range FAQ.

## Checklist

### DIFF
--- a/packages/document/main-doc/docs/zh/guides/troubleshooting/builder.mdx
+++ b/packages/document/main-doc/docs/zh/guides/troubleshooting/builder.mdx
@@ -566,7 +566,7 @@ Less 中除法的写法也可以通过配置项来修改，详见 [Less - Math](
 
 从 MAJOR_VERSION.47.0 版本开始，Modern.js 优化了 Type Checker 的检查范围。在之前的版本中，Type Checker 只输出 src 目录的类型错误，导致其他目录的类型错误无法被正确输出。
 
-在新版本中，Modern.js 的 Type Checker 对齐了原生 tsc 的类型检查范围，能够完整输出项目中的类型错误，你也可以通过 tsconfig.json 的 `include` 和 `exclude` 字段来配置类型检查范围。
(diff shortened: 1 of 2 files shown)


packages/document/main-doc/docs/zh/guides/troubleshooting/builder.mdx
```diff
@@ -562,6 +562,28 @@ Less 中除法的写法也可以通过配置项来修改，详见 [Less - Math](
 
 ---
 
+### 升级 Modern.js 版本后，检查出新的 TypeScript 类型错误？
+
+从 MAJOR_VERSION.47.0 版本开始，Modern.js 优化了 Type Checker 的检查范围。在之前的版本中，Type Checker 只输出 src 目录的类型错误，导致其他目录的类型错误无法被正确输出。
+
+在新版本中，Modern.js 的 Type Checker 对齐了原生 tsc 的类型检查范围，能够完整输出项目中的类型错误，你也可以通过 tsconfig.json 的 `include` 和 `exclude` 字段来配置类型检查范围。
```
SoonIter, line 569: `include` and `exclude` fields in tsconfig.json are different from `include` in `fork-ts-checker`

Fork-ts-checker issue.include filters the results and errors of tsc **output**.

tsconfig.json `include` and `exclude` is to config the tsc **input** (tsc actually check all files in the input and the all files within input files `imports`.)

:p If it were me, I would recommend filter output ... Tsc is so magical and hard to control.

SoonIter (approved):

packages/document/main-doc/docs/zh/guides/troubleshooting/builder.mdx
chenjiahan: I think user should know that both `tsconfig.include/exclude` and `tsChecker.include/exclude` will affect the type checker.

`tsconfig.include/exclude` should be configured correctly first, as it will affect the build performance.

The document has been updated to be more precise.

SoonIter (approved):

PR #5550, opened 2024-03-19
chore(deps): bump rsbuild 0.5.x

9aoy:
## Summary

reference: https://github.com/web-infra-dev/rsbuild/releases/tag/v0.5.0

## Checklist


packages/cli/uni-builder/src/shared/parseCommonConfig.ts
```diff
@@ -263,6 +272,30 @@ export async function parseCommonConfig(
...
+          if (isFunction(tag)) {
+            // The function will be executed at the end of the HTML processing flow
+            handlers.push(tag);
+          } else {
+            tags.push(tag);
+          }
+        }
+
+        return handlers.reduce(
+          (currentTags, handler) => handler(currentTags, utils) || currentTags,
```
chenjiahan, line 292: it is unexpected to run the handlers inside the uni-builder.

uni-builder just need to:

- transform `html.tagsByEntries` to `html.tags`
- sort the `html.tags` array, put functions at last

9aoy: we can't sort and put functions directly, because we should filter `entries` by `utils.entryName` first, which can only be geted in `HtmlInjectTagHandler`, and `HtmlInjectTagHandler` does not support returning functions/arrays.

chenjiahan (approved):

PR #5561, opened 2024-03-21
fix: dev proxy can not validate

GiveMe-A-Name:
## Checklist

### DIFF
--- a/packages/server/server/src/createDevServer.ts
+++ b/packages/server/server/src/createDevServer.ts
@@ -16,7 +16,6 @@ import {
   startWatcher,
   onRepack,
   getDevOptions,
-  transformToRsbuildServerOptions,
   initFileReader,
 } from './helpers';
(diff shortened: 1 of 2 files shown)


yimingjfe (approved):

packages/server/server/src/createDevServer.ts
```diff
@@ -61,13 +60,10 @@ export const createDevServer = async <O extends ServerBaseOptions>(
     close,
     onUpgrade,
   } = (await getMiddlewares?.({
-    ...transformToRsbuildServerOptions(dev),
     compress:
```
9aoy, line 63: It looks that `feat-new-server`branch lags behind the `main` branch. The main branch already uses rsbuild 0.5.x, and the code here is slightly different from before.

https://github.com/web-infra-dev/modern.js/pull/5550/files#diff-7d7530ee958c6d949a528b1db7ef0b3a3aec6ee3b2d63bc970719b293b9a3a7a

PR #5565, opened 2024-03-21
docs: Update c03-css.mdx

tho-us-if:
It's been written as shown in lines
154. mkdir -p src/components/Item
155. touch src/components/Item/index.tsx

but in line 301 while referring to the above mentioned file it's denoted as src/routes/components/Item/index.tsx -- which doesn't exist if one follows through the tutorial.

## Checklist


chenjiahan:
Nice catch 👍

PR #5567, opened 2024-03-22
fix: we should override server.baseUrl when merge it

GiveMe-A-Name:
## Summary

We should overrides `server.baseUrl` when user set it as `['/a', '/b']`.

## Checklist


GiveMe-A-Name:
!canary

zllkjc (approved):

PR #5606, opened 2024-04-01
fix: remove node: prefix

GiveMe-A-Name:
## Summary

We should support node v14.17.6

But node v14.17.6 don't support require('node:xx')

## Checklist

### DIFF
--- a/.changeset/dull-melons-tickle.md
+++ b/.changeset/dull-melons-tickle.md
@@ -1,0 +1,6 @@
+---
+'@modern-js/utils': patch
+---
+
+fix: remove `node:` prefix
+fix: 移除 `node:` 前缀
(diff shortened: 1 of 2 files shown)


yimingjfe (approved):

chenjiahan:
Do we still need to support Node 14?

GiveMe-A-Name:
> Do we still need to support Node 14?

1.  We have some user using node v14.
2. Our home page announce that "Before getting started, you will need to install [Node.js](https://nodejs.org/), and ensure that your **Node.js version is higher than 14.17.6**. We recommend using the LTS version of Node.js 18."

So I think we have better support node v14.
In future one days, we don't need support it, but is not now.

PR #5612, opened 2024-04-02
fix: unit test

GiveMe-A-Name:
## Checklist


packages/runtime/plugin-runtime/tests/ssr/serverRender/renderToStream/index.test.ts
```diff
@@ -73,7 +73,7 @@ afterEach(() => {
 
 describe('renderToStream', () => {
   test('basic usage', async () => {
-    const pipe = await render({ App, context, config });
+    const readable = (await render({ App, context, config })) as Readable;
```
yimingjfe, line 76: Can the render function not return the correct type?

yimingjfe (approved):

PR #5642, opened 2024-04-15
fix(uni-builder): modernjs should extend "ua" type in output.polyfill

fi3ework:
## Summary

Related to https://github.com/web-infra-dev/rsbuild/issues/2091.

## Checklist

### DIFF
--- a/packages/cli/uni-builder/src/types.ts
+++ b/packages/cli/uni-builder/src/types.ts
@@ -14,6 +14,7 @@ import type {
   RsbuildPluginAPI,
   ArrayOrNot,
   HtmlTagDescriptor,
+  Polyfill,
 } from '@rsbuild/shared';
 import type { RsbuildConfig } from '@rsbuild/core';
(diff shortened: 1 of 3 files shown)


packages/cli/uni-builder/src/types.ts
```diff
@@ -380,7 +381,9 @@ export type UniBuilderPlugin = {
 export type UniBuilderConfig = {
   dev?: RsbuildConfig['dev'];
   html?: RsbuildConfig['html'];
-  output?: RsbuildConfig['output'];
+  output?: Omit<NonNullable<RsbuildConfig['output']>, 'polyfill'> & {
```
chenjiahan, line 384: I think we also need to convert the `ua` value to `off`, to make sure that no polyfill will be injected by Rsbuild

fi3ework: Nice catch, done.

chenjiahan (approved):

PR #5697, opened 2024-04-28
fix: <Html> should add <html> props types

GiveMe-A-Name:
## Checklist


clChenLiang (approved):
LGTM

PR #5717, opened 2024-05-08
fix: server support add extends res headers from routes

GiveMe-A-Name:
## Checklist

### DIFF
--- a/packages/server/core/src/base/middlewares/renderHandler/render.ts
+++ b/packages/server/core/src/base/middlewares/renderHandler/render.ts
@@ -104,27 +104,20 @@ export async function createRender({
       metrics,
     };
 
-    let response: Response | void;
-
     switch (renderMode) {
       case 'data':
-        response = await dataHandler(req, renderOptions);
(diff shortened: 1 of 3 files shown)


packages/server/core/src/base/middlewares/renderHandler/render.ts
```diff
@@ -104,21 +104,27 @@ export async function createRender({
...
       case 'ssr':
       case 'csr':
-        return renderHandler(req, renderOptions, renderMode, onError);
+        response = await renderHandler(req, renderOptions, renderMode, onError);
+        break;
       default:
         throw new Error(`Unknown render mode: ${renderMode}`);
     }
+
+    applyExtendHeaders(response, routeInfo);
```
yimingjfe, line 125: Should the data handler's response be extended with headers?

GiveMe-A-Name: done~

yimingjfe (approved):

PR #5718, opened 2024-05-08
feat(server-core): render support dynamic route

GiveMe-A-Name:
## Checklist

### DIFF
--- a/packages/server/core/src/base/middlewares/renderHandler/render.ts
+++ b/packages/server/core/src/base/middlewares/renderHandler/render.ts
@@ -1,6 +1,8 @@
 import type { IncomingMessage } from 'http';
 import { Logger, Metrics, Reporter, ServerRoute } from '@modern-js/types';
 import { cutNameByHyphen } from '@modern-js/utils/universal';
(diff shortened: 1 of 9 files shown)


packages/server/core/src/base/middlewares/renderHandler/render.ts
```diff
@@ -36,6 +38,35 @@ interface CreateRenderOptions {
...
+    const urlPath = originUrlPath.endsWith('/')
+      ? `${originUrlPath}*`
+      : `${originUrlPath}/*`;
+    router.add('*', urlPath, route);
+  }
+
+  return router;
+}
+
+function matchRoute(
```
yimingjfe, line 58: In the public middleware, there is also a matchRoute, should this one be used as well?

GiveMe-A-Name: No, public middleware don't need handle dynamic params.
On the other hand, public middleware matchRoutes is not same as render middleware's

yimingjfe (approved):

PR #5719, opened 2024-05-08
feat(devtools): create new storage preset by remote function

Asuka109:
## Summary

- Create new storage preset item and write it to config file.
- Watch on the changes of config files and syncing it to client.

## Checklist

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -83,8 +83,5 @@
     "**/.DS_Store": true,
     "**/Thumbs.db": true,
     "**/node_modules": false
-  },
-  "[javascript]": {
-    "editor.defaultFormatter": "esbenp.prettier-vscode"
   }
 }
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -83,5 +83,8 @@
     "**/.DS_Store": true,
     "**/Thumbs.db": true,
     "**/node_modules": false
+  },
+  "[javascript]": {
+    "editor.defaultFormatter": "esbenp.prettier-vscode"
   }
 }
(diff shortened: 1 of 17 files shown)


.vscode/settings.json
```diff
@@ -83,5 +83,8 @@
     "**/.DS_Store": true,
     "**/Thumbs.db": true,
     "**/node_modules": false
+  },
+  "[javascript]": {
+    "editor.defaultFormatter": "esbenp.prettier-vscode"
```
zllkjc, line 88: What's the purpose of this change. any docs ?

packages/devtools/plugin/src/cli.ts
```diff
@@ -50,7 +52,27 @@ export const devtoolsPlugin = (
...
-        prepare: rpc.hooks.prepare,
+        async prepare() {
+          await rpc.hooks.prepare();
+
+          const appConfig = api.useConfigContext();
+          const appCtx = api.useAppContext();
+          const { devtools: options = {} } = appConfig;
+          updateContext(ctx, options);
+
+          const watcher = chokidar.watch(getConfigFilenames(), {
```
zllkjc, line 63: do we need to close in restart hook

.vscode/settings.json
Asuka109: Seems a unexpected change by vscode itself. I will revert it.

packages/devtools/plugin/src/cli.ts
Asuka109: Sure, I will do it.

Asuka109:
@zllkjc Done there~

zllkjc (approved):

PR #5739, opened 2024-05-14
fix(server-core): support custom server middleware redirect

GiveMe-A-Name:
## Checklist

### DIFF
--- a/packages/server/core/src/base/middlewares/customServer/index.ts
+++ b/packages/server/core/src/base/middlewares/customServer/index.ts
@@ -12,7 +12,7 @@ import {
   createCustomMiddlewaresCtx,
   createAfterStreamingRenderContext,
 } from './context';
-import { createBaseHookContext } from './base';
+import { ResArgs, createBaseHookContext } from './base';
 
 // eslint-disable-next-line @typescript-eslint/no-empty-function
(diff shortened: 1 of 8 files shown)


packages/server/core/src/base/middlewares/customServer/index.ts
```diff
@@ -168,7 +168,15 @@ export class CustomServer {
...
 
-      const customMiddlewareCtx = createCustomMiddlewaresCtx(c, locals);
+      const resArgs: ResArgs = {
+        headers: new Headers(),
+      };
+
+      const customMiddlewareCtx = createCustomMiddlewaresCtx(
+        c,
+        locals,
+        resArgs,
```
yimingjfe, line 178: It may not be a reasonable approach to get the value of a private variable using an external variable

GiveMe-A-Name: I think have no other ways to implement it.
Have any suggestions?

yimingjfe: Conclusion: we have some ways to modify this, but the server hook will be deprecated, so we don't modify it.

yimingjfe (approved):

PR #5752, opened 2024-05-16
feat: support deploy command

yimingjfe:
## Checklist


zllkjc:
!canary

packages/solutions/app-tools/src/plugins/deploy/index.ts
```diff
@@ -0,0 +1,68 @@
...
+    );
+  }
+
+  return createPreset(appContext, modernConfig, needModernServer);
+}
+
+export default (): CliPlugin<AppTools> => ({
+  name: '@modern-js/plugin-deploy',
+  setup: api => {
+    const deployTarget = process.env.MODERNJS_DEPLOY || provider || 'node';
```
GiveMe-A-Name, line 49: What  may be the value of provider?

GiveMe-A-Name:
LGTM

packages/solutions/app-tools/src/plugins/deploy/index.ts
yimingjfe: Currently, the value is "vercel" or "netlify".

GiveMe-A-Name: Got it

packages/solutions/app-tools/src/plugins/deploy/platforms/netlify.ts
```diff
@@ -0,0 +1,143 @@
...
+        await fse.copy(distDirectory, funcsDirectory, {
+          filter: (src: string) => {
+            const distStaticDirectory = path.join(distDirectory, `static`);
+            return !src.includes(distStaticDirectory);
+          },
+        });
+      }
+
+      // redirect files don't need to be in the function directory
+      const redirectFilePath = path.join(distDirectory, '_redirects');
```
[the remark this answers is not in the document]
zllkjc: And can developers continue to use netlify.toml in project ? like add headers for some route.
The `remix netlify adapter` seems to generate a toml file in the root directory, and use user's toml by default

packages/solutions/app-tools/src/plugins/deploy/platforms/node.ts
```diff
@@ -0,0 +1,84 @@
...
+  const entryFilePath = path.join(outputDirectory, 'index.js');
+  return {
+    async prepare() {
+      await fse.remove(outputDirectory);
+    },
+    async writeOutput() {
+      await fse.copy(distDirectory, outputDirectory);
+    },
+    async genEntry() {
+      if (!needModernServer) {
```
zllkjc, line 31: now csr project will also print then content in the console....but actually there is no .output/index.js file
img_v3_02b5_8100b41d-7f00-4fc7-8a08-6eb41c2fbf3g

zllkjc: we should fix it in `end()` hook

packages/solutions/app-tools/src/plugins/deploy/utils.ts
```diff
@@ -0,0 +1,72 @@
...
+    lambdaDirectory,
+    metaName,
+  } = appContext;
+  return {
+    sharedDirectory: `path.join(__dirname, "${path.relative(
+      appDirectory,
+      sharedDirectory,
+    )}")`,
+    apiDirectory: `path.join(__dirname, "${path.relative(
+      appDirectory,
```
zllkjc, line 27: emmm, can we just use `.` or '/'

packages/server/prod-server/src/netlify.ts
```diff
@@ -0,0 +1,19 @@
+import { loadServerEnv } from '@modern-js/server-core/base/node';
+import { createServerBase } from '@modern-js/server-core/base';
+import { initProdMiddlewares } from './init';
+import { BaseEnv, ProdServerOptions } from './types';
+
+export { initProdMiddlewares, type InitProdMiddlewares } from './init';
+
+export type { ProdServerOptions, BaseEnv } from './types';
+
+export const createNetlifyFunction = async (options: ProdServerOptions) => {
```
zllkjc, line 10: Why add it here instead of in the netlify entry ?

packages/solutions/app-tools/src/plugins/deploy/platforms/netlifyEntry.js
zllkjc, line 1: why use js file..

packages/solutions/app-tools/src/plugins/deploy/dependencies/index.ts
```diff
@@ -0,0 +1,262 @@
...
+  resolveTracedPath,
+} from './utils';
+
+export const handleDependencies = async (
+  appDir: string,
+  serverRootDir: string,
+  include: string[],
+) => {
+  const base = '/';
+  const entryFiles = await findEntryFiles(serverRootDir);
```
zllkjc, line 24: It seems static/ files are also in the directory, should it be 'entry' ?

packages/solutions/app-tools/src/plugins/deploy/dependencies/index.ts
zllkjc, line 1: In the future, we may be able to package this feature as a separate tool.
like rslog....https://web-bnpm.byted.org/package/rslog

zllkjc:
img_v3_02b5_36d3e239-7f9b-4a31-9517-65f43d407feg
can we give a better ui for this...at least use ipv4 ?

zllkjc:
add some docs impove in https://github.com/yimingjfe/modern.js/pull/1 @yimingjfe ~

packages/solutions/app-tools/src/plugins/deploy/index.ts
```diff
@@ -0,0 +1,68 @@
...
+      async beforeDeploy() {
+        const appContext = api.useAppContext();
+        const modernConfig = api.useResolvedConfigContext();
+        const deployPreset = await getDeployPreset(
+          appContext,
+          modernConfig,
+          deployTarget as DeployTarget,
+        );
+
+        deployPreset?.prepare && (await deployPreset?.prepare());
```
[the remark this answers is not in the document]
yimingjfe: > maybe we should use Object-oriented to develop it hhh

Is it similar to java? I think it's because java isn't flexible enough that we need to write it this way, but we're using JS =、=

packages/solutions/app-tools/src/plugins/deploy/platforms/netlify.ts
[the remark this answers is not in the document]
yimingjfe: > should we add docs for this file and static file for netlify

This file is hosted by framework and does not need to be understood by users.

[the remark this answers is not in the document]
yimingjfe: > And can developers continue to use netlify.toml in project ? like add headers for some route.
The remix netlify adapter seems to generate a toml file in the root directory, and use user's toml by default

Yeah, Developers continue to use thier netlify.toml in project,  includes writing their own redirects and headers.

For the developer's own netlify.toml,  it's mentioned in the document,  and it's necessary.

packages/solutions/app-tools/src/plugins/deploy/index.ts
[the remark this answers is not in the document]
yimingjfe: > And shouldn't we directly write the deploy code in the deploy() function in apptools main process, instead of using plugins and writing it in the hook?

As shown in the code, the hooks needed for deployment are fairly simple, the plugin hooks are more complex, and I didn't want to add some hooks to the plugin core and complicate things in the first version.

packages/solutions/app-tools/src/plugins/deploy/platforms/node.ts
yimingjfe: This is expected, when it is a Node.js environment, users needs Modern.js Server to serve the web page.

packages/server/prod-server/src/netlify.ts
yimingjfe: There is no substitute for Netlify entry, which uses this

packages/solutions/app-tools/src/plugins/deploy/platforms/netlifyEntry.js
yimingjfe: It's a template file

yimingjfe:
> img_v3_02b5_36d3e239-7f9b-4a31-9517-65f43d407feg can we give a better ui for this...at least use ipv4 ?

This is the server in the production environment. We should use ipv6. I think the address of ipv6 can be exposed. It is clear, and other frameworks do the same.

packages/solutions/app-tools/src/plugins/deploy/dependencies/index.ts
yimingjfe: Yes, there's a problem here in the node.js scenario, let me fix it

packages/solutions/app-tools/src/plugins/deploy/dependencies/index.ts
yimingjfe: Yeah, I think it may be more appropriate as a bundler plugin.

packages/solutions/app-tools/src/plugins/deploy/platforms/node.ts
yimingjfe: Got it, done, in the CSR, the file is also emitted

packages/solutions/app-tools/src/plugins/deploy/dependencies/index.ts
yimingjfe: done~

tests/tsconfig.json
```diff
@@ -4,5 +4,6 @@
     "strictNullChecks": false,
     "noEmit": true
   },
-  "types": ["jasmine", "jest"]
+  "types": ["jasmine", "jest"],
+  "exclude": ["./integration"]
```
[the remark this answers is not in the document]
yimingjfe: done~

packages/solutions/app-tools/src/plugins/deploy/index.ts
yimingjfe: The value is injected by `std-env`, which is a package imported by framework

packages/solutions/app-tools/src/plugins/deploy/dependencies/index.ts
yimingjfe: I may have misunderstood your means, indeed, the feature may be available on some scenarios, such as for server-side frameworks, like `Nest.js`, `Artus.js`

packages/solutions/app-tools/src/plugins/deploy/index.ts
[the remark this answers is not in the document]
zllkjc: > As shown in the code, the hooks needed for deployment are fairly simple, the plugin hooks are more complex, and I didn't want to add some hooks to the plugin core and complicate things in the first version.

I have encountered a problem before, which is that the plugin needs to control the order of itself and `appTools`. This often leads to some hooks needing to be executed before `appTools` and some hooks needing to be executed later, So it cannot be implemented. I think this is a problem with our hook and internal logic design.

For the current code, I think `deploy` is our core logic and should not be executed in beforeDeploy..Maybe the better way is:
```
runner.beforeDeploy();
deploy(); // or runner.deploy()
runner.afterDeploy();

```

packages/solutions/app-tools/src/plugins/deploy/platforms/netlify.ts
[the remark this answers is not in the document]
zllkjc: Ok, so I have anthor question.

netlify use `headers` as follow:  
https://docs.netlify.com/configure-builds/file-based-configuration/#headers

```toml
[[headers]]
  for = "/*"
```

How do our developers write the `for` field ?

in remix, it generate a default configuration:
```toml
[[headers]]
for = "/build/"

[headers.values]
````

packages/solutions/app-tools/src/plugins/deploy/utils.ts
zllkjc: I'm not sure we talk the same thing. I mean the code:
`${path.relative(appDirectory, appDirectory)`

zllkjc:
> This is the server in the production environment, we should support dual stack and I think the address of ipv6 can be exposed. because It is clear, and other frameworks do the same.

Ok, whether it is necessary to give the user an accessible ipv4 address here..if user start in local, then he can just click the url and preview the page.

packages/solutions/app-tools/src/plugins/deploy/platforms/netlifyEntry.js
zllkjc: So we can't use `.ts` to write it ? Because in the end we still use the js file, which is JS, are you worried that the code will not be pure after compiling?

packages/solutions/app-tools/src/plugins/deploy/index.ts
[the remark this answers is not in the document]
yimingjfe: done, add a new `deploy` hook to plugin core

packages/solutions/app-tools/src/plugins/deploy/platforms/netlifyEntry.js
yimingjfe: Yeah, I  think it's not necessary to write `ts`, the code is simple, the output of compiling may not be as expected, placeholders will have type errors and pure js is better for find the problem

yimingjfe:
> > This is the server in the production environment, we should support dual stack and I think the address of ipv6 can be exposed. because It is clear, and other frameworks do the same.
> 
> Ok, whether it is necessary to give the user an accessible ipv4 address here..if user start in local, then he can just click the url and preview the page.

This address can accept both ipv6 and ipv4 connections, including `localhost`, `127.0.0.1`

Are you finding it inappropriate to display `[::]`?  I'm guessing it might be a matter of habit? We could listen to `[::]` and show the ipv4 address, but I don't think that's clear

GiveMe-A-Name (approved):

PR #5800, opened 2024-06-04
feat(devtools): hook system for devtools plugin

Asuka109:
## Checklist


zllkjc:
It seems that most of the code is same as before, and now it has become very clear~

zllkjc (approved):

PR #5820, opened 2024-06-12
feat(ssr): support for loaderFailureMode configure

yimingjfe:
## Checklist


packages/cli/plugin-data-loader/src/cli/createRequest.ts
```diff
@@ -49,6 +49,20 @@ const handleDeferredResponse = async (res: Response) => {
   return res;
 };
 
+const isErrorResponse = (res: Response) => {
+  return res.headers.get('X-Modernjs-Error') != null;
+};
+
+const handleErrorResponse = async (res: Response) => {
+  if (isErrorResponse(res)) {
+    const data = await res.json();
```
GiveMe-A-Name, line 58: Does may Response's body be a can't-JSON object?

yimingjfe: The res is an ErrorResponse only if body is json format，you can refer to the following code:
1. https://github.com/yimingjfe/modern.js/blob/8ce4e8d10d1deb02a8db670fc1c0a5e461c6d96f/packages/cli/plugin-data-loader/src/runtime/index.ts#L153

2. https://github.com/yimingjfe/modern.js/blob/8ce4e8d10d1deb02a8db670fc1c0a5e461c6d96f/packages/cli/plugin-data-loader/src/runtime/errors.ts#L142

GiveMe-A-Name (approved):

PR #5849, opened 2024-06-19
feat: enable stage3 decorator for swc based on typescript version

### LINKED ISSUE #5834: [Feature]: build stage 3 decorator with swc in modern.js module

I'd like to build a module which uses stage3 decorator. In modern.js app + rspack, I can easily configure swc to do so:
``` ts
export default defineConfig<'rspack'>({
  plugins: [
    appTools({
      bundler: 'experimental-rspack',
    }),
  ],
  tools: {
    swc: {
      jsc: {
        transform: {
          decoratorVersion: '2022-03',
        },
      },
    },
  },
});
```

However, it seems modern.js module is using esbuild and there is no way to explicitly configure using swc or enable swc for this decorator transform.

Example file you can test with:
``` ts

function loggedMethod<This, Args extends any[], Return>(
  target: (this: This, ...args: Args) => Return,
  context: ClassMethodDecoratorContext<
    This,
    (this: This, ...args: Args) => Return
  >,
) {
  const methodName = String(context.name);

  function replacementMethod(this: This, ...args: Args): Return {
    console.log(`LOG: Entering method '${methodName}'.`);
    const result = target.call(this, ...args);
    console.log(`LOG: Exiting method '${methodName}'.`);
    return result;
  }

  return replacementMethod;
}

export class Person {
  name: string;
  constructor(name: string) {
    this.name = name;
  }

  @loggedMethod
  greet() {
    console.log(`Hello, my name is ${this.name}.`);
  }
}
```

Maybe use SWC when "experimentalDecorator" of tsconfig is true?

Related: https://modernjs.dev/module-tools/en/guide/advance/in-depth-about-build.html#use-swc

yf-yang:
## Summary

This is a draft PR as the API is not designed. I'd like to discuss with the maintenance team if this implementation is fine.

Currently, I can hack it by enabling `transformLodash` to enable swc compiler.

## Related Links

Fixes #5834 
## Checklist

### DIFF
--- a/packages/solutions/module-tools/src/builder/feature/swc.ts
+++ b/packages/solutions/module-tools/src/builder/feature/swc.ts
@@ -40,7 +40,7 @@ const getSwcTarget = (target: string): JscTarget => {
   return 'es2022';
 };
 
-export const swcTransform = (userTsconfig: ITsconfig) => ({
+export const swcTransform = (userTsconfig: ITsconfig, tsVersion: number) => ({
   name,
   apply(compiler: ICompiler) {
     const tsUseDefineForClassFields =
@@ -102,6 +102,10 @@ export const swcTransform = (userTsconfig: ITsconfig) => ({
                 useDefineForClassFields,
                 legacyDecorator: emitDecoratorMetadata ? true : undefined,
                 decoratorMetadata: emitDecoratorMetadata ? true : undefined,
+                decoratorVersion:
+                  !emitDecoratorMetadata && tsVersion >= 5
+                    ? '2022-03'
+                    : undefined,
               },
               externalHelpers,
               target: getSwcTarget(target),
--- a/packages/solutions/module-tools/src/builder/feature/index.ts
+++ b/packages/solutions/module-tools/src/builder/feature/index.ts
@@ -1,5 +1,5 @@
 import { HookList, Context } from '../../types';
-import { getProjectTsconfig } from '../../utils/dts';
+import { detectTSVersion, getProjectTsconfig } from '../../utils/dts';
 import { formatCjs } from './format-cjs';
 import { css } from './style';
 import { minify } from './terser';
@@ -29,6 +29,10 @@ export async function getInternalList(context: Context): Promise<HookList> {
   const emitDecoratorMetadata =
     userTsconfig?.compilerOptions?.emitDecoratorMetadata ?? false;
 
+  const tsVersion = await detectTSVersion(
+    context.api.useAppContext().appDirectory,
+  );
+
   const { transformImport, transformLodash, externalHelpers, format, target } =
     context.config;
 
@@ -42,7 +46,7 @@ export async function getInternalList(context: Context): Promise<HookList> {
     : format === 'umd' || target === 'es5';
   if (enbaleSwcTransform) {
     const { swcTransform } = await import('./swc');
-    internal.push(swcTransform(userTsconfig));
+    internal.push(swcTransform(userTsconfig, tsVersion));
   }
 
   if (enableSwcRenderChunk) {


packages/solutions/module-tools/src/builder/feature/swc.ts
```diff
@@ -40,7 +40,7 @@ const getSwcTarget = (target: string): JscTarget => {
   return 'es2022';
 };
 
-export const swcTransform = (userTsconfig: ITsconfig) => ({
+export const swcTransform = (userTsconfig: ITsconfig, tsVersion: number) => ({
```
yf-yang, line 43: Is it OK to add a parameter like that?

packages/solutions/module-tools/src/builder/feature/swc.ts
```diff
@@ -102,6 +102,10 @@ export const swcTransform = (userTsconfig: ITsconfig) => ({
                 useDefineForClassFields,
                 legacyDecorator: emitDecoratorMetadata ? true : undefined,
                 decoratorMetadata: emitDecoratorMetadata ? true : undefined,
+                decoratorVersion:
+                  !emitDecoratorMetadata && tsVersion >= 5
+                    ? '2022-03'
+                    : undefined,
```
fi3ework, line 108: This could be a breaking change for projects using ts^5 and `emitDecoratorMetadata` not enabled?

packages/solutions/module-tools/src/builder/feature/swc.ts
fi3ework: This function is an internal, seems acceptable.

packages/solutions/module-tools/src/builder/feature/swc.ts
yf-yang: AFAIK this option takes effect only when decorator grammar is used?

yf-yang:
What's your opinion on the configuration option?
Option 1: An option to enable SWC.
Option 2: An option to enable decorator. Then implicitly enable SWC and checks ts version to apply the correct SWC config.

packages/solutions/module-tools/src/builder/feature/swc.ts
fi3ework: - Just took a look into the codebase, you're right that this PR won't break user when `emitDecoratorMetadata` is set. 
- However, as per https://github.com/web-infra-dev/modern.js/blob/main/packages/solutions/module-tools/src/builder/feature/index.ts#L35-L39, SWC transform is enabled only in certain conditions. New trigger condition is needed to enable the `'2022-03'` config.
- `tsVersion > 5` is not sufficient, user could still use legacy decorator with `experimentalDecorators` flag, https://www.typescriptlang.org/docs/handbook/release-notes/typescript-5-0.html#differences-with-experimental-legacy-decorators.
- esbuild start to support the stage3 decorator since https://github.com/evanw/esbuild/releases/v0.21.0 with heuristic legacy backward support. IMO, bumping esbuild to latest version could be a way more straightforward solution.

yf-yang: 🤔 I'm fine with either. If you are willing to add another trigger option I'd like to help implement it. If you can bump to latest esbuild that'd be great 😀

fi3ework: I'd prefer bumping esbuild as we should catch up the latest version and the feature is resolved by the way. We're now using `0.19.2` version of esbuild which is release 10 months ago. It's better to bump it, @chenjiahan is there any issue blocks bumping esbuild that I don't know?

fi3ework: Talked with @chenjiahan, there's no blocking point stopping us from bumping esbuild. We'd inclined to bump esbuild to address this issue.
I'm closing this PR in favor of bumping esbuild. Let me know if you're interested in implement this. Thanks for the PR anyway.

PR #5876, opened 2024-06-25
feat(devtools/client): switch to hash router

Asuka109:
## Summary

- Switch the router of @modern-js/devtools-client to hash router, so that the client can be deployed with any pathname.
- Expose `createManifest` hook to enable plugins modify the manifest.
- Initialize React DevTools renderer on `plugin/src/runtime.ts`.
  - Fix crashing when left and visit the DevTools tab again.
  - Take controll from React DevTools extension.

## Checklist


packages/devtools/plugin/src/runtime.ts
```diff
@@ -54,10 +82,12 @@ const setup = () => {
   const cookieManifestUrl = document.cookie.match(
     /use_modernjs_devtools=([^;]+)/,
   )?.[1];
-  const devtoolsUrl = cookieManifestUrl || injectedManifestUrl;
+  const devtoolsUrl = injectedManifestUrl || cookieManifestUrl;
```
zllkjc, line 85: What is the reason for the logical change here?

Asuka109: For some cases developers may serve both development and production server at a same domain.  For the previous implementation, if people setup a devtools cookie for production service, it will also affect the development server. So it has being changed to use the manifest url injected by bundler in primary.

zllkjc (approved):

zllkjc (approved):

PR #5911, opened 2024-07-05
docs: update client overlay default value

9aoy:
## Checklist


Timeless0911 (approved):

Timeless0911:
Does en doc need to be changed?

9aoy:
> Does en doc need to be changed?

No need, the default value in the en doc is correct.

PR #5919, opened 2024-07-09
fix: should use server-entires error & modify ServerRoot as ReactElement

GiveMe-A-Name:
## Checklist

### DIFF
--- a/packages/runtime/plugin-runtime/src/core/server/requestHandler.ts
+++ b/packages/runtime/plugin-runtime/src/core/server/requestHandler.ts
@@ -26,7 +26,7 @@ export type HandleRequestOptions = Exclude<
 
 export type HandleRequest = (
   request: Request,
-  ServerRoot: React.ReactElement, // App, routes,
+  ServerRoot: React.ComponentType, // App, routes,
   options: HandleRequestOptions,
 ) => Promise<Response>;
 
@@ -127,7 +127,7 @@ function createSSRContext(
 export const createRequestHandler: CreateRequestHandler =
   async handleRequest => {
     const requestHandler: RequestHandler = async (request, options) => {
-      const serverRoot = createRoot();
+      const Root = createRoot();
 
       const runner = getGlobalRunner();
 
@@ -214,9 +214,11 @@ export const createRequestHandler: CreateRequestHandler =
         `${CHUNK_CSS_PLACEHOLDER}</head>`,
       );
 
-      const ServerRoot = React.createElement(serverRoot, {
-        _internal_context: Object.assign(context, { ssr: true }),
-      });
+      const ServerRoot = (props: any) =>
+        React.createElement(Root, {
+          ...props,
+          _internal_context: Object.assign(context, { ssr: true }),
+        });
 
       const response = await handleRequest(request, ServerRoot, {
         ...options,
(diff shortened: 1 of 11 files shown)


packages/runtime/plugin-runtime/src/core/server/requestHandler.ts
```diff
@@ -25,7 +26,7 @@ export type HandleRequestOptions = Exclude<
 
 export type HandleRequest = (
   request: Request,
-  serverRoot: React.ComponentType, // App, routes,
+  ServerRoot: React.ReactElement, // App, routes,
```
zllkjc, line 29: the name for `ReactElement` type is better use lowercase, serverRoot is good

packages/runtime/plugin-runtime/src/core/server/requestHandler.ts
```diff
@@ -213,7 +214,11 @@ export const createRequestHandler: CreateRequestHandler =
         `${CHUNK_CSS_PLACEHOLDER}</head>`,
       );
 
-      const response = await handleRequest(request, serverRoot, {
+      const ServerRoot = React.createElement(serverRoot, {
```
zllkjc, line 217: We should pass a variable capable of writing JSX syntax to the `requestHandler`, then user can write `<ServerRoot>` in their code. So, should it be written like this here:

```tsx
const ServerRoot = (props: any) => {
  return React.createElement(createRoot(), {
    ...props,
    _internal_context: Object.assign(context, { ssr: true }),
  });
};
```

packages/runtime/plugin-runtime/src/core/server/shared.ts
```diff
@@ -7,13 +7,13 @@ export type SSRConfig = NonNullable<ServerUserConfig['ssr']>;
 
 export type RenderStreaming = (
   request: Request,
-  serverRoot: React.ComponentType<any>,
+  ServerRoot: React.ReactElement,
```
zllkjc, line 10: also serverRoot is better

packages/runtime/plugin-runtime/src/core/server/string/index.ts
zllkjc, line 1: 这个文件也是，应该不用改造，renderString 的时候，拿到的是 React Element，用小写会好

zllkjc:
I think we should modify the file `template.server.ts`, like: 

```tsx
const handleRequest = async (request, ServerRoot, options) => {

  const body = await #render(request, <ServerRoot />, options);

  return new Response(body, {
    headers: {
      'content-type': 'text/html; charset=utf-8',
      #headers
    },
  })
};

```

packages/runtime/plugin-runtime/src/core/server/requestHandler.ts
GiveMe-A-Name: Done

packages/runtime/plugin-runtime/src/core/server/shared.ts
GiveMe-A-Name: Done~

packages/runtime/plugin-runtime/src/core/server/string/index.ts
GiveMe-A-Name: done~

packages/runtime/plugin-runtime/src/core/server/requestHandler.ts
GiveMe-A-Name: Done~

zllkjc (approved):

PR #5925, opened 2024-07-09
feat: add server monitors interface

GiveMe-A-Name:
## Summary
- unity modern monitors interface
- provide `monitors` to extend logger, timing.

## Checklist


zllkjc (approved):

packages/server/core/src/plugins/render/index.ts
```diff
@@ -133,8 +134,9 @@ function createRenderHandler(
     const nodeReq = c.env.node?.req;
 
     const res = await render(request, {
-      logger,
       nodeReq,
+      monitors,
+      logger,
```
yimingjfe, line 139: Why does the render function need to take logger and metrics as inputs?

GiveMe-A-Name: For void break change, we have export logger & metrics api to user by using `useRuntimeContext`

zllkjc (approved):

PR #5937, opened 2024-07-11
refactor(devtools): service worker

Asuka109:
## Summary

image

- Refactor the service worker plugin: `packages/devtools/client/plugins/framework/service-worker.ts`
- Service worker persist data via IndexDB (with unstorage API) insteading of url queries.
- Service worker handle requests via Hono.
- Implement RPC between application and service worker via Hono.
- Add some toasts.
- Fix some styling issues.

## Checklist


packages/devtools/client/src/utils/service-agent.ts
```diff
@@ -1,12 +1,32 @@
...
 }
 
-export interface ServiceStatus {
-  href: string;
-  version: string;
-  rules: ModifyHeaderRule[];
-}
+export const SERVICE_SCRIPT = '/sw-devtools.js';
+
+export const swClient = hc<AppType>('https://__devtools-api.modernjs.dev');
```
zllkjc, line 13: `https://__devtools-api.modernjs.dev` is also used in `service-worker.ts`, we can use a constant

PR #5940, opened 2024-07-11
refactor: adapt Rsbuild environment config

9aoy:
## Summary

bump rsbuild 1.0.0-alpha and use Rsbuild environment config
-  adapt Rsbuild environment config
-  compat serviceWorker environment in uni-builder
-  compat dev.startUrl / dev.beforeStartUrl / distPath.server / distPath.worker in uni-builder
-  support html minify and compat tools.htmlPlugin minify option
-  remove entry and target Options in createUniBuilder

## Related Links

https://github.com/web-infra-dev/rsbuild/discussions/2508

## Checklist


packages/cli/uni-builder/src/shared/plugins/htmlMinify.ts
chenjiahan, line 1: What's the difference between this plugin and https://github.com/rspack-contrib/rsbuild-plugin-html-minifier-terser

packages/cli/uni-builder/tests/__snapshots__/styledComponents.test.ts.snap
```diff
@@ -176,7 +178,7 @@ exports[`plugins/styled-components > should enable ssr when target contain node
               "<WORKSPACE>/node_modules/<PNPM_INNER>/babel-plugin-import/lib/index.js",
               {
                 "camel2DashComponentName": false,
-                "libraryDirectory": "lib",
+                "libraryDirectory": "es",
```
chenjiahan, line 181: should this be `lib` when target is node?

packages/cli/uni-builder/src/shared/plugins/htmlMinify.ts
9aoy: Support applying `tools.htmlPlugin` minify option, and `tools.htmlPlugin` supports updating minifyConfig based on entryName, which is difficult to support in `rsbuild-plugin-html-minifier-terser`.

packages/cli/uni-builder/tests/__snapshots__/styledComponents.test.ts.snap
9aoy: This is the snapshot of target web. Previously, when the rsbuild target had node, all library directories used lib, but it seems to me that only node should use lib and web should use es.

image

PR #5973, opened 2024-07-19
refactor: dev-server

GiveMe-A-Name:
## Summary
- refactor dev-server's mock middleware
- `@modern-js/app-tools` will createDevServer from '@modern-js/server' instead of 'uni-builder'
- tools.devServer after, before, `setupMiddlewares` will apply mock middleware before and after

## Checklist


packages/server/server/src/helpers/mock.ts
```diff
@@ -67,64 +76,93 @@ export const registerMockHandlers = async ({
...
+    return false;
+  });
+
+  return matched;
+};
+
+export async function initOrUpdateMockMiddlewares(pwd: string) {
+  const mockModule = await getMockModule(pwd);
+
+  const mockHandlers = mockModule?.default;
```
yimingjfe, line 126: In the getMockModule function, it also require `module.default`, so does this need to get default property twice?

packages/server/server/src/helpers/mock.ts
```diff
@@ -67,64 +76,93 @@ export const registerMockHandlers = async ({
...
   }
 
-  if (!mockModule) {
+  if (!mockHandlers) {
     throw new Error(`Mock file ${mockFilePath} parsed failed!`);
   }
 
-  Object.entries(mockModule as MockConfig).forEach(([key, handler]) => {
+  return {
+    default: mockHandlers,
```
yimingjfe, line 100: Maybe `mockHandlers` is better than `default`?

yimingjfe (approved):

packages/server/server/src/helpers/mock.ts
GiveMe-A-Name: Yes, Great idea

GiveMe-A-Name: Done～

yimingjfe (approved):

PR #5983, opened 2024-07-23
feat(ssg): support ssg with bff.enableHandleWeb

yimingjfe:
## Checklist

### DIFF
--- a/packages/cli/plugin-ssg/src/index.ts
+++ b/packages/cli/plugin-ssg/src/index.ts
@@ -47,7 +47,7 @@ export const ssgPlugin = (): CliPlugin<AppTools> => ({
         const routes = readJSONSpec(buildDir);
 
         // filter all routes not web
-        const pageRoutes = routes.filter(route => !route.isApi);
+        const pageRoutes = routes.filter(route => !route.isApi || route.isSPA);
         const apiRoutes = routes.filter(route => route.isApi);
 
         // if no web page route, skip ssg render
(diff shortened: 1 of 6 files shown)


packages/runtime/plugin-runtime/src/core/server/utils.ts
```diff
@@ -64,17 +64,12 @@ export function getSSRConfigByEntry(
...
-  if (staticGenerate) {
-    return true;
-  }
-
   if (ssrByEntries?.[entryName]) {
     return ssrByEntries[entryName];
   }
 
-  return ssr!;
+  return ssr || true;
```
zllkjc, line 72: this change will cause all csr project run in ssr mode ?

packages/cli/plugin-ssg/src/index.ts
```diff
@@ -47,7 +47,7 @@ export const ssgPlugin = (): CliPlugin<AppTools> => ({
         const routes = readJSONSpec(buildDir);
 
         // filter all routes not web
-        const pageRoutes = routes.filter(route => !route.isApi);
+        const pageRoutes = routes.filter(route => !route.isApi || route.isSPA);
```
zllkjc, line 50: in the end of this function, there is a line `writeJSONSpec(buildDir, pageRoutes.concat(apiRoutes));`

which may cause duplicate routes...and `createServer` also use `apiRoutes`

zllkjc: It can work normally, but it's not elegant haha

yimingjfe: Yeah, it's not elegant here, because if we want it to be relatively elegant, we either need to change the information in route.json or need to make a lot of changes to this file.

Do you have any ideas?

yimingjfe: If I create a new object to merge the information of pageRoutes and apiRoutes as the data source of the page to be rendered and route.json, do you think this is appropriate?

zllkjc: can we juest use `route.isSPA` to judge ？

zllkjc (approved):

PR #5984, opened 2024-07-23
docs: update html-rspack-plugin link

9aoy:
## Checklist

### DIFF
--- a/packages/document/builder-doc/docs/en/config/tools/htmlPlugin.md
+++ b/packages/document/builder-doc/docs/en/config/tools/htmlPlugin.md
@@ -25,7 +25,7 @@ const defaultHtmlPluginOptions = {
 };
 ```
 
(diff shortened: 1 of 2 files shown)


packages/document/builder-doc/docs/en/config/tools/htmlPlugin.md
chenjiahan, line 1: Can we just link to Rsbuild documentation?

9aoy: rsbuild's `tools.htmlPlugin` documentation lacks the HTML Minification functionality and related defaults, but modern.js does. I'm worried that the link will confuse users.

chenjiahan: Just link to Rsbuild and add some extra contents about HTML modification to Modern.js doc.

chenjiahan (approved):

PR #5992, opened 2024-07-24
fix: we should generate bootstrap.server.jsx when ssr mode

GiveMe-A-Name:
## Checklist


2heal1 (approved):

2heal1:
that's last piece of mf rspack ssr  !

PR #6031, opened 2024-08-01
feat(module-tools): add resolve.alias, to support redirect module id

10Derozan:
## Summary
When we use `alias: { "react": "react-native"}` and want to redirect the module id, it failed.
We will get a result: "./react-native", but we want "react-native",
Now we use a new config, which names "resolve.alias" to support it.
In this config, we remove the default alias, and make it easier to understand.

## Checklist


10Derozan:
If it have be approved, I will add a document to introduce it.

jkzing (commented):
We need to make sure that the behavior remains the same as before when using the `alias` configuration. In the current implementation, `resolve.alias` and `alias` have the same behavior, this should be incorrect.

Timeless0911 (approved):

PR #6075, opened 2024-08-14
fix: add the user-defined `output.assetPrefix` config for the node mode of modern deploy

Gk0Wk:
## Summary

When using `modern deploy`, the project's configured `output.assetPrefix` is ignored, which results in static resource files failing to be accessed in the Node.js deployment mode.

For example, if the user configures `assetPrefix` as `/foo`, the URLs for the generated static resource files will start with `/foo/static`. However, if this configuration is ignored in the generated `.output/index.js`, the production server will continue to expect static resources to be prefixed with `/static`, leading to failures in accessing the static resource files.

## Checklist

### DIFF
--- a/.changeset/shy-dingos-crash.md
+++ b/.changeset/shy-dingos-crash.md
@@ -1,0 +1,5 @@
+---
+'@modern-js/app-tools': major
+---
+
+fix: add a user-defined output.assetPrefix path for the node mode of modern deploy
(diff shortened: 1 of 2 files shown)


yimingjfe:
Thank you for your PR. Have you encountered this problem in a real project before? Because Modern.js will serialize the `modern.config.ts` to server and I also didn't reproduce the problem.

PR #6095, opened 2024-08-19
docs(main-doc): add swc and esbuild plugin

9aoy:
## Summary

fix link broken. swc and esbuild plugin are only used in modern.js, and modern.js builder is no longer maintained.

## Checklist


packages/document/main-doc/docs/en/configure/app/tools/esbuild.mdx
chenjiahan, line 1: Can we remove these docs from builder doc?

chenjiahan (approved):

PR #6111, opened 2024-08-21
fix: only inline runtimeChunk/script/style in production

9aoy:
## Summary

Be consistent with documentation. `disableInlineRuntimeChunk` / `enableInlineScripts` / `enableInlineStyles` configurations only worked  in production mode.

image

## Checklist


9aoy:
need fix in rsbuild

PR #6151, opened 2024-09-02
fix: we should run runner.fallback when ssr fallback

GiveMe-A-Name:
## Checklist


spencerHT (approved):
LGTM

PR #6163, opened 2024-09-03
fix(server): avoid the require problem on windows

yimingjfe:
## Checklist

### DIFF
--- a/packages/server/core/src/plugins/render/render.ts
+++ b/packages/server/core/src/plugins/render/render.ts
@@ -255,18 +255,22 @@ async function renderHandler(
   const { serverManifest } = options;
 
   const ssrByRouteIds = options.config.server?.ssrByRouteIds;
+  const runtimeEnv = getRuntimeEnv();
 
   if (
     serverManifest.nestedRoutesJson &&
     ssrByRouteIds &&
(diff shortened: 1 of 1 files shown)


packages/server/core/src/plugins/render/render.ts
```diff
@@ -255,18 +255,22 @@ async function renderHandler(
...
+    ssrByRouteIds?.length > 0 &&
+    runtimeEnv === 'node'
   ) {
     const { nestedRoutesJson } = serverManifest;
     const routes = nestedRoutesJson?.[options.routeInfo.entryName!];
-    const runtimeEnv = getRuntimeEnv();
-    if (routes && runtimeEnv === 'node') {
+
+    if (routes) {
+      const { pathToFileURL } = await import('node:url');
```
GiveMe-A-Name, line 270: The node:url would be bundle-in when esbuild build web target bundle.

GiveMe-A-Name: Can we import like this: 
```ts
const urlPath = 'node:url';
const { pathToFileURL } = await import(urlPath);
```

GiveMe-A-Name (approved):

PR #6165, opened 2024-09-03
fix(server): avoid possible memory leaks

yimingjfe:
## Checklist


spencerHT (approved):
LGTM

PR #6192, opened 2024-09-09
fix: react's hook useId cause ssr hydrate failed

GiveMe-A-Name:
## Summary

- Add some empty node to sure the browser hydrate has same struct with ssr render.

## Checklist

### DIFF
--- a/.changeset/strong-hounds-attend.md
+++ b/.changeset/strong-hounds-attend.md
@@ -1,0 +1,6 @@
+---
+'@modern-js/runtime': patch
+---
+
+fix: react's hook useId cause ssr hydrate failed
+fix: react's hook useId 导致 ssr hydrate 失败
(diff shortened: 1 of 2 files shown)


yimingjfe:
Is there more information to this solution? Just looking at the code, I don't really understand why it can be solved  the problem.

GiveMe-A-Name:
> Is there more information to this solution? Just looking at the code, I don't really understand why it can solved the problem.

React would generate the id reference the node tree. In router/runtime/plugin.node.ts, we organize the component like this: 
```ts
// https://github.com/web-infra-dev/modern.js/blob/v2.59.0/packages/runtime/plugin-
return (
                <>
                  <StaticRouterProvider
                    router={remixRouter!}
                    context={routerContext!}
                    hydrate={false}
                  />

                  {mode === 'stream' && (
                    // ROUTER_DATA will inject in `packages/runtime/plugin-runtime/src/core/server/string/ssrData.ts` in string ssr
                    // So we can inject it only when streaming ssr
                    <DeferredDataScripts
                      nonce={nonce}
                      context={routerContext!}
                    />
                  )}
                  {mode === 'stream' && JSX_SHELL_STREAM_END_MARK}
                </>
);
```
We must sure the browser as the same struct as ssr

yimingjfe (approved):

PR #6220, opened 2024-09-14
fix: router plugin baseUrl not work

caohuilin:
## Checklist


KyrieLii (approved):

KyrieLii:
LGTM, we need to explain the factors affecting basename in the document.
- server.baseUrl
- runtime.router.basename
- garfish plugin
- some inner logic
...

PR #6286, opened 2024-09-27
feat(deps): unpin Rsbuild patch version

9aoy:
## Summary

Unpin Rsbuild patch version, so users can upgrade to the latest patch version of rsbuild without upgrading Modern.js.

## Checklist


chenjiahan (approved):

GiancarlosIO:
Hi @chenjiahan 

This is great! thanks for that!!

Could you please help me to understand how can I upgrade to the latest patch version of rsbuild without upgrading Modern.js?

I will need to add/install rsbuild dependencies in my package.json? 🤔

chenjiahan:
If you are using pnpm, try https://pnpm.io/cli/update

Other package managers should have similar commands

PR #6342, opened 2024-10-11
fix(builder): `output.sourceMap` not work

SoonIter:
## Related Links

@9aoy

## Checklist

### DIFF
--- a/packages/cli/uni-builder/src/shared/parseCommonConfig.ts
+++ b/packages/cli/uni-builder/src/shared/parseCommonConfig.ts
@@ -95,6 +95,7 @@ export async function parseCommonConfig(
       enableAssetFallback,
       enableAssetManifest,
       disableSourceMap,
+      sourceMap,
       convertToRem,
       disableMinimize,
       polyfill,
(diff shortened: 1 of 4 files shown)


packages/cli/uni-builder/src/shared/parseCommonConfig.ts
```diff
@@ -326,6 +327,7 @@ export async function parseCommonConfig(
     pluginGlobalVars(globalVars),
     pluginDevtool({
       disableSourceMap,
+      sourceMap,
```
9aoy, line 330: If ﻿`output.sourcemap.js` is define, I think we do not need to apply ﻿`pluginDevtool`.

SoonIter: if disableSourceMap and sourceMap are set both, what behavior should we do

9aoy: `output.sourceMap`, because `output.disableSourceMap` is deprecated.

SoonIter: done

9aoy (approved):

PR #6382, opened 2024-10-14
fix(bff): zodError causing 500 error due to inconsistent zod versions

keepview:
## Checklist

### DIFF
--- a/.changeset/empty-cups-swim.md
+++ b/.changeset/empty-cups-swim.md
@@ -1,0 +1,5 @@
+---
+'@modern-js/bff-core': patch
+---
+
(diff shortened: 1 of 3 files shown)


.changeset/empty-cups-swim.md
```diff
@@ -0,0 +1,5 @@
+---
+'@modern-js/bff-core': patch
+---
+
+fix(bff): 修复 zod 版本不统一导致的 zodError 触发 500 错误码
```
yimingjfe, line 5: English changeset is also needed

yimingjfe:
Does the lock file need to be changed?

PR #6609, opened 2024-12-05
feat: support set tools.babel in environment config

9aoy:
## Summary
In some scenarios, different babel configs need to be set up in different environments because the webpack babel plugin is built into uni-builder, we need support set `tools.babel` in environment config.

## Checklist


packages/cli/uni-builder/src/rspack/index.ts
```diff
@@ -49,7 +50,47 @@ export async function parseConfig(
...
+        );
+      } else if (uniBuilderConfig.tools?.babel) {
+        environmentConfig.plugins ??= [];
+        environmentConfig.plugins.push(
+          pluginBabel({
+            babelLoaderOptions: uniBuilderConfig.tools?.babel,
+          }),
+          pluginBabelPost(),
+        );
+      }
```
chenjiahan, line 91: These two conditions can be merged to reduce duplicated code.

chenjiahan (approved):

PR #6612, opened 2024-12-05
feat: support automatic jsx runtime while using react@>=16.14.0

Asuka109:
## Summary

https://github.com/facebook/react/releases/tag/v16.14.0

## Checklist


Asuka109:
@9aoy @zllkjc The previous implementation of initializating @babel/preset-react configuration only enable new jsx runtime for react with version >=17. But it should be applied for >=16.14.0 too. Otherwise it will prevent to use some components with new runtime.

zllkjc (approved):

9aoy (approved):

PR #6641, opened 2024-12-18
fix: compat jest plugin hooks

caohuilin:
## Summary

Adds compatibility for Jest plugins hooks.

Due to the presence of multiple parameters in the jestConfig hook, the `transformHookParams` function add 
`isMultiple` flag to indicate whether the current parameter is multiple, ensuring correct function execution.

## Checklist


packages/solutions/app-tools/src/compat/index.ts
```diff
@@ -35,6 +40,8 @@ export const compatPlugin = (): CliPluginFuture<AppTools<'shared'>> => ({
   },
   registryHooks: {
     appendEntryCode: createCollectAsyncHook<AppendEntryCodeFn>(),
+    jestConfig: createAsyncHook<JestConfigFn>(),
+    afterHest: createAsyncHook<AfterTestFn>(),
```
zllkjc, line 44: typo 'Hest' ?

zllkjc (approved):

PR #6669, opened 2024-12-26
feat(cssnano): disable `normalizeUrl` while enabling `output.injectStyles`

Asuka109:
## Summary

The `normalizeUrl` will transform relative url from `./assets/img.svg` to `assets/img.svg`.
It may break the behavior of webpack resolver while using style-loader.
So disable it while `output.injectStyles = true` or `output.disableCssExtract = true`.

## Checklist


9aoy (approved):

packages/cli/uni-builder/src/shared/plugins/postcss.ts
```diff
@@ -25,6 +30,24 @@ export const pluginPostcss = ({
...
+            // merge longhand will break safe-area-inset-top, so disable it
+            // https://github.com/cssnano/cssnano/issues/803
+            // https://github.com/cssnano/cssnano/issues/967
+            mergeLonghand: false,
+            /**
+             * normalizeUrl will transform relative url from `./assets/img.svg` to `assets/img.svg`.
+             * It may break the behavior of webpack resolver while using style-loader.
+             * So disable it while `output.injectStyles = true` or `output.disableCssExtract = true`.
+             */
+            normalizeUrl: !injectStyles,
```
9aoy, line 46: `injectStyles` has already been checked in `enableCssMinify` and cssnano is only enabled when enableCssMinify, it seems to be repeated here.

Asuka109: Got it, so I can just turn off normalizeUrl in all cases right?

9aoy: I think so.

9aoy (approved):

PR #6670, opened 2024-12-26
fix: `postcss.config.*` file not working

9aoy:
## Summary

Fix `postcss.config` not working, use postcss function configuration instead of object configuration, object configuration will overwrite user's `postcss.config`.

https://rsbuild.dev/config/tools/postcss#object-type

## Checklist


packages/cli/uni-builder/src/shared/plugins/postcss.ts
```diff
@@ -61,10 +61,19 @@ export const pluginPostcss = ({
...
         {
           tools: {
-            postcss: {
-              postcssOptions: {
-                plugins,
-              },
+            postcss: opts => {
+              if (typeof opts.postcssOptions === 'object') {
+                opts.postcssOptions.plugins!.push(...plugins);
+                return;
```
chenjiahan, line 67: unnecessary return

packages/cli/uni-builder/src/shared/plugins/postcss.ts
```diff
@@ -61,10 +61,19 @@ export const pluginPostcss = ({
...
+              } else if (typeof opts.postcssOptions === 'function') {
```
chenjiahan, line 68: How about using `addPlugins` method? It can be simpler.

packages/cli/uni-builder/tests/__snapshots__/postcssLegacy.test.ts.snap
```diff
@@ -143,55 +143,6 @@ exports[`plugin-postcssLegacy > should allow tools.postcss to override the plugi
           "postcssOptions": {
             "config": false,
             "plugins": [
-              {
-                "Once": [Function],
```
chenjiahan, line 147: Why are these snapshots been removed

9aoy: There was a problem with this case before. The user's `tools.postcss` object configuration should override the default plugins.

packages/cli/uni-builder/src/shared/plugins/postcss.ts
9aoy: `addPlugins` is added as extraPlugins and cannot be modified or overridden by the user

packages/cli/uni-builder/package.json
```diff
@@ -73,6 +73,7 @@
     "postcss-media-minmax": "5.0.0",
     "postcss-nesting": "12.1.5",
     "postcss-page-break": "3.0.4",
+    "postcss-load-config": "6.0.1",
```
chenjiahan, line 76: I guess we should use v5 because v6 requires Node 18

image

tests/integration/tailwindcss/fixtures/tailwindcss-without-plugin/modern.config.ts
chenjiahan, line 1: Can we create a `postcss` folder and add more test cases for different usages?

- Merge PostCSS plugins
- Read postcss.config.*
- Usage with `@modern-js/plugin-tailwindcss`

chenjiahan (approved):

PR #6703, opened 2025-01-07
 feat: deprecate beforeConfig hook

caohuilin:
## Checklist

### DIFF
--- a/packages/solutions/app-tools/src/compat/hooks.ts
+++ b/packages/solutions/app-tools/src/compat/hooks.ts
@@ -28,9 +28,6 @@ export function getHookRunners(
     /**
      * app tools hooks
      */
-    beforeConfig: async () => {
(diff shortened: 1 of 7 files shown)


packages/solutions/app-tools/src/compat/hooks.ts
```diff
@@ -202,18 +199,20 @@ export function handleSetupResult(
...
-            params,
-          );
-          if (isMultiple) {
-            return transformHookResult(key, await fn(...transformParams));
-          } else {
-            return transformHookResult(key, await fn(transformParams));
-          }
-        });
+      if (newAPI) {
+        if (api[newAPI]) {
```
zllkjc, line 203: We can use `newApi && api[newApi]` instead

zllkjc (approved):

PR #6774, opened 2025-01-26
feat: load env in appTools when before the first time load modern config

zllkjc:
## Summary

For some historical reasons, now Modern.js will load config twice when use exec any command. But we will only load env file one time. So in the first time we load config, the env not existed, this will cause some problems.

```ts
// modern.config.ts
console.log(process.env.MY_ENV)
```

```ts
// .env
MY_ENV=modern.js
```

In this PR, we run `loadEnv()` in `isAutoLoadPlugin()` to ensure load env before every time we load config.

> In next major version, we will remove `isAutoLoadPlugin`, so that config file and env file will only load one time.

## Checklist


packages/solutions/app-tools/src/run/index.ts
```diff
@@ -74,7 +74,9 @@ export async function run({
   const finalConfigFile = customConfigFile || getConfigFile(configFile);
   const autoLoadPlugins = await isAutoLoadPlugins(
     appDirectory,
-    finalConfigFile,
+    finalConfigFile || 'modern.config.ts',
```
caohuilin, line 77: The default value of modern.config.ts is not needed? It is already included in finalConfigFile.

zllkjc: Done~

caohuilin (approved):

giancarlosisasi:
OMG thanks @zllkjc !

Last week I was blocked from upgrading to the latest version because, maybe for the behavior fixed in this pr, the environment variables weren't working.

After testing, 2.64.1 now it works without errors. Thanks!

For more context, in the `modern.config.ts` we are running a function to validate that some required env variables has the correct value:

```tsx
// always validate that all required env vars are defined before building the app
validateThatAllEnvVarsRequiredAreDefined();

export const createModernConfig = ({
....
```

PR #6799, opened 2025-02-06
fix(dev-server): apply dev configuration, avoid dev.setupMiddleware not apply.

GiveMe-A-Name:
## Summary
Modern.js has `dev` and `tools.devServer` configuration. And `tools.devServer` already deprecated, we should using `dev` config instead firstly.

## Checklist

### DIFF
--- a/packages/solutions/app-tools/src/commands/dev.ts
+++ b/packages/solutions/app-tools/src/commands/dev.ts
@@ -89,20 +89,24 @@ export const dev = async (
 
   const pluginInstances = await loadServerPlugins(api, appDirectory, metaName);
 
+  const toolsDevServerConfig = normalizedConfig.tools.devServer;
+
   const serverOptions = {
     metaName,
     dev: {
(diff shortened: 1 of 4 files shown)


caohuilin (approved):

packages/solutions/app-tools/src/commands/dev.ts
```diff
@@ -92,10 +92,17 @@ export const dev = async (
...
+      devMiddleware: {
+        writeToDisk: normalizedConfig.dev.writeToDisk,
+      },
       port,
-      https: normalizedConfig.dev.https,
       host: normalizedConfig.dev.host,
-      ...normalizedConfig.tools?.devServer,
+      https: normalizedConfig.dev.https,
+      hot: normalizedConfig.dev.hmr,
+      setupMiddlewares: normalizedConfig.dev.setupMiddlewares,
```
zllkjc, line 105: if `dev.setupMiddlewares` or `dev.writeToDisk` is undefined, will it overrides the old value ?

GiveMe-A-Name: Got it

GiveMe-A-Name: Done~

zllkjc (approved):

PR #6877, opened 2025-02-23
fix(uni-builder): fix compilation error when using lodash on windows devices

Forzoom:
## Summary
在 windows 设备上，当页面中引入 lodash 相关包时，构建将出错。

主要原因看起来是：babel-plugin-lodash 中通过 glob.sync 来查找 lodash 相关文件夹，但 [glob](https://www.npmjs.com/package/glob/v/9.3.5) 在某次更新后加入了 windowsPathsNoEscape 配置，对于 path.join 等方式生成的路径，需要配置 windowsPathsNoEscape: true，否则无法正常找到 lodash 相关文件，导致构建失败。

改动点：在内置的 compiled/babel-plugin-lodash 的 glob.sync 调用 options 中补充 windowsPathsNoEscape: true。

目前 windows 设备可以通过配置 transformLodash: false 来绕过该问题

## Related Links

repro: https://github.com/Forzoom/repro-mdjs-lodash-error

## Checklist


9aoy:
This bug was caused when we migrated `Modern.js builder` to `Rsbuild` + `uni-builder`, where we mistakenly upgraded the major version of `glob`.

The dependency issue in `babel-plugin-lodash` should be fixed by updating the prebundle script, similar to previous versions of Modern.js.  I will create a new PR to fix this. 🌹 

Reference https://github.com/web-infra-dev/modern.js/blob/v2.42.1/scripts/prebundle/src/constant.ts#L389

PR #7017, opened 2025-04-07
docs: update node version

caohuilin:
## Checklist

### DIFF
--- a/.changeset/light-seals-divide.md
+++ b/.changeset/light-seals-divide.md
@@ -2,6 +2,6 @@
 '@modern-js/main-doc': patch
 ---
 
-docs: Recommended Node version changed to 20, minimum supported Node version changed to 18
+docs: Recommended Node version changed to 22, minimum supported Node version changed to 18
 
-docs: 推荐使用 Node 版本修改为 20，最低支持 Node 版本修改为 18
+docs: 推荐使用 Node 版本修改为 22，最低支持 Node 版本修改为 18
--- a/.changeset/light-seals-divide.md
+++ b/.changeset/light-seals-divide.md
@@ -1,0 +1,7 @@
+---
+'@modern-js/main-doc': patch
+---
+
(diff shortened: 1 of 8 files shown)


zllkjc (approved):

giancarlosisasi:
Sorry to coment in a already-merged MR but @caohuilin do you think we can also update the docs fro vercel deployment? 🙏🏼 

https://github.com/web-infra-dev/modern.js/blob/main/packages/document/main-doc/docs/en/guides/basic-features/deploy.mdx#full-stack-project-1

It recommends node 18 but maybe we can use now node20 (that's the most recent version that vercel supports)

```
"engines": {
  "node": "20.x"
}
```

If you are ok with this, maybe @yimingjfe should be also aware of this because (as I noticed in the issue) he is fixing https://github.com/web-infra-dev/modern.js/issues/7022 which is also related to node versions 🙏🏼

caohuilin:
@giancarlosisasi Thank you for your feedback. I've received your response regarding the Node.js version update. I'll discuss this with @yimingjfe  to evaluate updating the Vercel deployment documentation from Node.js 18 to Node.js 20. We appreciate your suggestion and will address this documentation update soon.

yimingjfe:
> Sorry to coment in a already-merged MR but @caohuilin do you think we can also update the docs fro vercel deployment? 🙏🏼
> 
> https://github.com/web-infra-dev/modern.js/blob/main/packages/document/main-doc/docs/en/guides/basic-features/deploy.mdx#full-stack-project-1
> 
> It recommends node 18 but maybe we can use now node20 (that's the most recent version that vercel supports)
> 
> ```
> "engines": {
>   "node": "20.x"
> }
> ```
> 
> If you are ok with this, maybe @yimingjfe should be also aware of this because (as I noticed in the issue) he is fixing #7022 which is also related to node versions 🙏🏼

I replied to you in the issue，Modern.js doesn't have any limitations on this，you can use version 20 or even higher.

PR #7106, opened 2025-04-29
feat: router v7 plugin

yimingjfe:
## Summary

This PR provides a `@modern-js/plugin-router-v7` plugin to use react-router v7 in the current Modern.js.

## Checklist


packages/runtime/plugin-runtime/src/router/runtime/DeferredDataScripts.node.tsx
zllkjc, line 1: In fact, the only change in this file is the way to get activeDeferreds ?

packages/cli/plugin-data-loader/src/runtime/index.ts
```diff
@@ -1,4 +1,5 @@
 import { transformNestedRoutes } from '@modern-js/runtime-utils/browser';
+import { activeDeferreds } from '@modern-js/runtime-utils/browser';
```
zllkjc, line 2: seems not used ?

packages/toolkit/runtime-utils/src/browser/nestedRoutes.tsx
```diff
@@ -131,6 +139,25 @@ function createLoader(route: NestedRoute): LoaderFunction {
...
+        if (typeof document === 'undefined') {
+          activeDeferreds = getAsyncLocalStorage()?.useContext()
+            ?.activeDeferreds as Map<string, DeferredData>;
+        } else {
+          activeDeferreds = originalActiveDeferreds;
+        }
+        if (
+          res !== null &&
+          typeof res === 'object' &&
+          Object.getPrototypeOf(res) === Object.prototype
```
zllkjc, line 154: What is the meaning of this line, could you add some comment ?

zllkjc:
So it seems that the biggest difference is that deferData has been removed from the results of `queryRoute`, and we need to save this information ourselves ?

yimingjfe:
> So it seems that the biggest difference is that deferData has been removed from the results of `queryRoute`, and we need to save this information ourselves ?

Yeah, react-router no longer processes deferred data, so we create and save `activeDeferreds` ourselves.

packages/runtime/plugin-router-v7/src/cli/index.ts
```diff
@@ -0,0 +1,42 @@
...
+export const routerPlugin = (): CliPluginFuture<AppTools> => ({
+  name: '@modern-js/plugin-router-v7',
+  required: ['@modern-js/runtime'],
+  setup: api => {
+    api.config(() => {
+      const config = api.getNormalizedConfig();
+      const alias = config?.source?.alias;
+      if (typeof alias !== 'undefined') {
+        Object.keys(alias).forEach(key => {
+          if (key.includes('react-router')) {
```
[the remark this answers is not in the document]
yimingjfe: Not all includes `react-router` here will cause problems, so there is no return.

packages/runtime/plugin-runtime/src/router/runtime/DeferredDataScripts.node.tsx
yimingjfe: Yes

packages/toolkit/runtime-utils/src/browser/nestedRoutes.tsx
yimingjfe: This ensures that only pure objects will be processed, I move this part into a function.

zllkjc (approved):

PR #7107, opened 2025-04-30
fix(server): should sync all headers from req.headers

yimingjfe:
## Summary

Avoid the problem that rawHeaders can only obtain the original request header. For example, the request header added by the middleware will not be synchronized at present

## Checklist

### DIFF
--- a/packages/server/core/src/adapters/node/node.ts
+++ b/packages/server/core/src/adapters/node/node.ts
@@ -25,11 +25,22 @@ export const createWebRequest = (
   body?: BodyInit,
 ): Request => {
   const headerRecord: [string, string][] = [];
-  const len = req.rawHeaders.length;
-  for (let i = 0; i < len; i += 2) {
-    const key = req.rawHeaders[i];
-    if (!key.startsWith(':')) {
-      headerRecord.push([key, req.rawHeaders[i + 1]]);
+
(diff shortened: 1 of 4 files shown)


packages/server/core/src/adapters/node/node.ts
```diff
@@ -25,11 +25,22 @@ export const createWebRequest = (
...
-    const key = req.rawHeaders[i];
-    if (!key.startsWith(':')) {
-      headerRecord.push([key, req.rawHeaders[i + 1]]);
+
+  for (const [key, value] of Object.entries(req.headers)) {
+    if (key.startsWith(':')) {
+      continue;
+    }
+
+    if (Array.isArray(value)) {
```
keepview, line 34: Is this dealing with the problem of the same header key being overwritten? The original implementation seems to be fine as well

yimingjfe: Mainly in order to handle the following cases, we can set the request header like this:
```
req['setCookie'] = [
          'cookie1=value1; Path=/',
          'cookie2=value2; HttpOnly'
        ]
```

Because `req.rawHeaders` and `req.headers` are different data structures, so the implementation was changed

zllkjc (approved):

zllkjc (approved):

PR #7111, opened 2025-05-02
docs(bff): Update import path to match current package names

42tg:
## Summary
The Documentation [here](https://modernjs.dev/guides/advanced-features/bff/sdk.html) and [here](https://modernjs.dev/guides/advanced-features/bff/cross-project.html) 
are referencing to an unknown import, since the `@modern-js/runtime/bff` seems not to export it accordingly.

## Related Links

* https://modernjs.dev/guides/advanced-features/bff/sdk.html
* https://modernjs.dev/guides/advanced-features/bff/cross-project.html

## Checklist

### DIFF
--- a/packages/document/main-doc/docs/en/guides/advanced-features/bff/cross-project.mdx
+++ b/packages/document/main-doc/docs/en/guides/advanced-features/bff/cross-project.mdx
@@ -90,7 +90,7 @@ The `configure` function from `${package_name}/runtime` supports domain configur
 
 ```ts title="src/routes/page.tsx"
 import { configure } from '${package_name}/runtime';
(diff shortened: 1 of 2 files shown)


yimingjfe:
Thanks for your feedback, `@modern-js/runtime/bff` has an exported configure function, you can refer to this [example](https://github.com/modern-js-dev/modern.js/blob/7ffa4ea5f8ea16c40d6d3493d68b8ab7d85e81fc/tests/integration/bff-corss-project/bff-indep-client-app/src/base/App.tsx#L2)

I guess it may be caused by the lack of `/// <reference types='@modern-js/plugin-bff/types' />` in the `modern-app-env.d.ts` in the project，when BFF is turned on using the generator, this is written by default.

PR #7129, opened 2025-05-09
fix(doc): don't briefly show 404 when navigating

vegerot:
https://github.com/user-attachments/assets/808564b3-2400-49d0-84bc-db20b3c6087b

Summary:
Currently, clicking on links will briefly show a 404 page before being
redirected.  This is because we incorrectly set the default route to be zh
while the default language is en.  This commit fixes the issue by aligning the
default route and language.

Test Plan:
1. Go to homepage
2. Click "Introduction"
Before: Briefly see 404 before being redirected
After: See the correct page

### DIFF
--- a/packages/document/main-doc/src/i18n/index.ts
+++ b/packages/document/main-doc/src/i18n/index.ts
@@ -9,7 +9,7 @@ const translations = {
 
 export function useUrl(url: string) {
   const lang = useLang();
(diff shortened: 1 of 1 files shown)


caohuilin:
Thanks!

vegerot:
@caohuilin clicking on Google Search results will still show 404 
image

I'm hoping this is just because the site needs to be recrawled. If not, we may need to add canonical URLs to pages.

PR #7147, opened 2025-05-13
feat!(create): upgrade default browser targets

vegerot:
See discussion in https://github.com/web-infra-dev/rspack/pull/10290 for info

### DIFF
--- a/packages/generator/generators/mwa-generator/templates/base-template/.browserslistrc.handlebars
+++ b/packages/generator/generators/mwa-generator/templates/base-template/.browserslistrc.handlebars
@@ -1,5 +1,4 @@
-chrome >= 51
-edge >= 15
-firefox >= 54
-safari >= 10
-ios_saf >= 10
+last 2 versions
+> 0.2%
+Firefox ESR
+not dead


packages/document/main-doc/docs/en/guides/advanced-features/page-performance/optimize-bundle.mdx
```diff
@@ -46,16 +46,12 @@ See the [performance.bundleAnalyze](/configure/app/performance/bundle-analyze.ht
 
 Modern.js will compile the code according to the project's Browserslist config, and inject some Polyfills. If the project does not need to be compatible with legacy browsers, you can adjust the Browserslist and drop some legacy browsers, thereby reducing the compilation overhead on syntax and polyfill.
 
-Modern.js's default Browserslist config is:
+Modern.js's projects are created with the [default](https://github.com/browserslist/browserslist/blob/7b0ff670e997c53b13d44e2150ab09807a20ed6b/index.js#L477) Browserslist config.
 
-```js
-['> 0.01%', 'not dead', 'not op_mini all'];
```
vegerot, line 52: off-topic: this was wrong. Did we ever use this as a default?

vegerot: nvm. I see this is the default when not using a `.browserslistrc`

chenjiahan:
I think the default values ​​of browserslist provided by Modern.js and Rspack should be different. Modern.js should consider the compatibility requirements of enterprise users such as ByteDance.

@zllkjc might be able to help evaluate the current goals of Modern.js for browser compatibility.

vegerot:
I disagree in two places

1. **Modern**.js by default shouldn't be paranoid about enterprise compatibility.
2. The proposed defaults are sensible even for ByteDance projects like TikTok.

---

1. Modern.js should pick the sensible defaults that work for the vast majority of people.  We make it easy for people to customize the defaults to their own liking.
2. The current defaults are targeting browsers over _nine years old_.  According to [browser usage statistics](https://caniuse.com/usage-table), this covers **99.99%** of current internet users. This is plenty, even for TikTok.  We can support 99.999% of current internet users by adding support for IE9. Should we do that too?

Also, my earlier comment about Vite is useful here https://github.com/web-infra-dev/rspack/pull/10290#issuecomment-2863956828

zllkjc:
Here's my opinion:

1. In Modern.js, I didn't get enough feedback or discussions about browserslist. Next.js seems to [not be using dynamic browsers](https://nextjs.org/docs/architecture/supported-browsers). 
> But I think we can consider using dynamic versions in Modern.js to get better performance. Or there might be other advantages ? I don't think coverage is an advantage of the dynamic version.

2. Inside ByteDance, I think we can't do that. What we need to ensure first is the stability of our business. Not all projects have the same user base as TikTok, and the users targeted by different projects are also different. For example, Volcengine may target more users who use older versions of browsers, while dynamic versions may change the compatibility of the code without developers being aware of it.

vegerot:
> next.js

image

My hypothesis is that Next.js did the same thing as [Vite](https://github.com/web-infra-dev/rspack/pull/10290#issuecomment-2863956828). Claim to be "modern" but actually just picked the most modern and widely supported browsers when it was released.

Even still, their defaults are still much newer than ours.

>  Not all projects have the same user base

Exactly.  That's why it's the goal of Modern.js to have good defaults and make it easy for users to customize it to their liking.  Again, if we have to paranoidly support every use case should we make IE support default?  The current default goes back to **2016**.

vegerot:
By the way, can we please acknowledge that the _current_ behavior when devs decide not to use a `.browserslistrc` is already what I'm proposing?

vegerot:
Agreed!

_New_ ByteDance projects by _default_ can support widely available browsers. Again, we're talking about browsers that **99.5%** of people use.  If there's a new project that needs to support a ten-year old browser they can easily update the config to do that.

zllkjc:
This PR is being closed for now, as we've completed a round of upgrades in the major version.

PR #7155, opened 2025-05-15
fix: ignore domain when detect whether the request is for static asset

zllkjc:
## Summary

This PR resolves the issue where static assets cannot be accessed when developers configure the current domain name as `assetPrefix`.

```ts
import { appTools, defineConfig } from "@modern-js/app-tools";

export default defineConfig({
  runtime: {
    router: true,
  },
  server: {
    port: 3052,
  },
  output: {
    assetPrefix:"http://localhost:3052"
  },
  plugins: [
    appTools({
      bundler: "rspack",
    }),
  ],
});
```

Because middleware also valida host, which is actually unnecessary since resource requests from non-current hosts cannot enter the server.

## Checklist

### DIFF
--- a/packages/server/core/src/adapters/node/plugins/static.ts
+++ b/packages/server/core/src/adapters/node/plugins/static.ts
@@ -161,8 +161,6 @@ export function createStaticMiddleware(
     // exist is path
     const hit = staticPathRegExp.test(pathname);
 
-    console.log('hit', hit, staticPathRegExp, pathname);
-
     // FIXME: shoudn't hit, when cssPath, jsPath, mediaPath as '.'
     if (hit) {
       const filepath = path.join(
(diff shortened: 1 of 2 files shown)


packages/server/core/src/adapters/node/plugins/static.ts
```diff
@@ -143,11 +161,13 @@ export function createStaticMiddleware(
     // exist is path
     const hit = staticPathRegExp.test(pathname);
 
+    console.log('hit', hit, staticPathRegExp, pathname);
```
keepview: remove？

zllkjc: done

yimingjfe (approved):

PR #7264, opened 2025-06-25
feat: server config hot reload

keepview:
## Checklist

### DIFF
--- a/packages/solutions/app-tools/src/commands/dev.ts
+++ b/packages/solutions/app-tools/src/commands/dev.ts
@@ -150,8 +150,6 @@ export const dev = async (
         );
       },
     );
-    setServer(server);
-    setServerOptions(options);
   } else {
     const options = {
       ...serverOptions,
@@ -177,7 +175,5 @@ export const dev = async (
         await afterListen();
       },
     );
-    setServerOptions(options);
-    setServer(server);
   }
 };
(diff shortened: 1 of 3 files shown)


packages/solutions/app-tools/src/commands/dev.ts
```diff
@@ -10,7 +10,11 @@ import {
 } from '@modern-js/utils';
 import type { AppNormalizedConfig, AppTools } from '../types';
 import { buildServerConfig } from '../utils/config';
-import { setServer } from '../utils/createServer';
+import {
+  createServer,
+  setServer,
+  setServerOptions,
```
zllkjc, line 16: `setServer` and `setServerOptions` is never used

packages/solutions/app-tools/src/plugins/serverRuntime.ts
```diff
@@ -12,5 +12,11 @@ export default (): CliPluginFuture<AppTools<'shared'>> => ({
         ],
       },
     }));
+    api._internalServerPlugins(({ plugins }) => {
```
zllkjc, line 15: Better create a new server plugin.

zllkjc: Btw, this plugin should not be used in production

packages/solutions/app-tools/src/utils/createServer.ts
```diff
@@ -1,33 +1,158 @@
...
+      console.error('Error closing existing server:', error);
+    }
+  }
+
+  const { server: newServer, afterListen } = await createDevServer(
+    options,
+    applyPluginsFn || applyPlugins,
+  );
+
+  server = newServer;
```
zllkjc, line 89: same  logic as `setServer`

PR #7321, opened 2025-07-10
fix: test case for mergeConfig change

caohuilin:
## Checklist

### DIFF
--- a/packages/toolkit/plugin-v2/src/cli/run/utils/mergeConfig.ts
+++ b/packages/toolkit/plugin-v2/src/cli/run/utils/mergeConfig.ts
@@ -1,11 +1,5 @@
-import { isOverriddenConfigKey } from '@modern-js/utils';
-import {
-  isArray,
-  isEqual,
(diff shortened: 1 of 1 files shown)


packages/toolkit/plugin-v2/src/cli/run/utils/mergeConfig.ts
```diff
@@ -32,12 +26,22 @@ export const mergeConfig = <Config, NormalizedConfig>(
...
+      const targetArray = ensureArray(target);
+      const sourceArray = ensureArray(source);
+      const allItems = [...targetArray, ...sourceArray];
+      const seenNonFunc: any[] = [];
+      const result: any[] = [];
+      for (const item of allItems) {
+        if (isFunction(item)) {
+          result.push(item);
+        } else {
+          if (!seenNonFunc.some(seen => isEqual(seen, item))) {
```
clChenLiang, line 38: why not use find?

PR #7339, opened 2025-07-11
feat: server config hot reload

keepview:
## Checklist

### DIFF
--- a/packages/solutions/app-tools/src/utils/createServer.ts
+++ b/packages/solutions/app-tools/src/utils/createServer.ts
@@ -2,11 +2,13 @@ import type { Server } from 'node:http';
 import type { Http2SecureServer } from 'node:http2';
 import { applyPlugins } from '@modern-js/prod-server';
 import {
+  type ApplyPlugins,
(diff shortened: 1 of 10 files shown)


packages/solutions/app-tools/src/utils/createServer.ts
```diff
@@ -15,3 +22,24 @@ export const closeServer = async () => {
...
+  afterListen: () => Promise<void>;
+}> => {
+  if (server) {
+    server.close();
+  }
+  const {
+    server: newServer,
+    afterListen,
+    reload: reloadDevServer,
+  } = await createDevServer(options, applyPluginsFn || applyPlugins);
```
zllkjc: We can pass server plugins in here directly, Do we still need to add an export?

zllkjc: And why not just put the plugin in `@modern-js/server`

keepview: done

zllkjc (dismissed):

PR #7371, opened 2025-07-16
refactor: remove styledComponent plugin from uni-builder

GiveMe-A-Name:
## Summary

Remove styledComponent plugin from uni-builder.
Recommend using [@rsbuild/plugin-styled-components](https://www.npmjs.com/package/@rsbuild/plugin-styled-components) to implement styled-component compile

## Checklist


packages/cli/uni-builder/tests/__snapshots__/default.test.ts.snap
chenjiahan, line 1: This PR adds some new snapshots. Is this the expected change?

GiveMe-A-Name: Updated. I merged the main branch to solved conflict.The added snapshots may imports in the merge action.
Surprisingly, vitest pass all snapshot tests. Maybe we can optimize this case in rstest.

chenjiahan: get 👌

PR #7404, opened 2025-07-21
feat: add dev.lazyCompilation instead of experiments.lazyCompilation

zllkjc:
## Summary

LazyCompilation is a stable features in Rspack, so move it to `dev.lazyCompilation`.

And we remove the part of Webpack. Now this configuration is the same as Rsbuild.

## Checklist

### DIFF
--- a/packages/document/main-doc/docs/en/configure/app/dev/lazy-compilation.mdx
+++ b/packages/document/main-doc/docs/en/configure/app/dev/lazy-compilation.mdx
@@ -11,23 +11,23 @@ type LazyCompilationOptions =
   | boolean
   | {
       /**
-       * 为 entries 启用 lazy compilation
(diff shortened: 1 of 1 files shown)


packages/document/main-doc/docs/en/configure/app/dev/lazy-compilation.mdx
```diff
@@ -0,0 +1,45 @@
...
+# dev.lazyCompilation
+
+- **Type:**
+
+```ts
+type LazyCompilationOptions =
+  | boolean
+  | {
+      /**
+       * 为 entries 启用 lazy compilation
```
chenjiahan: should be English here

zllkjc: done!

PR #7405, opened 2025-07-21
feat(bff): should not compress the api response

yimingjfe:
## Summary

When the streaming response is returned, if it is compressed, the streaming fails, here we make the bff's response not to be streamed.

## Checklist

### DIFF
--- a/packages/cli/plugin-bff/src/cli.ts
+++ b/packages/cli/plugin-bff/src/cli.ts
@@ -1,3 +1,4 @@
+import type { IncomingMessage } from 'http';
 import path from 'path';
 import type { AppTools, CliPlugin } from '@modern-js/app-tools';
 import { ApiRouter } from '@modern-js/bff-core';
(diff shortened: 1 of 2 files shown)


zllkjc (approved):

packages/cli/plugin-bff/src/cli.ts
```diff
@@ -144,6 +145,25 @@ export const bffPlugin = (): CliPlugin<AppTools> => ({
...
+        if (
+          typeof devServer === 'object' &&
+          devServer !== null &&
+          !Array.isArray(devServer)
+        ) {
+          const compress = devServer.compress;
+          if (typeof compress === 'undefined' || compress === true) {
+            devServer.compress = {
+              filter: (req: IncomingMessage) => {
+                return !req.url?.includes(prefix);
```
tmkx, line 161: it be more appropriate to use `.startsWith` here?

PR #7415, opened 2025-07-24
refactor: rename @modern-js/uni-builder as @modern-js/builder

GiveMe-A-Name:
## Summary

@modern-js will remove webpack in next major version. @modern-js/uni-builder would not have any about webpack.
So we rename @modern-js/uni-builder as @modern-js/builder

## Checklist


chenjiahan (approved):
👍

PR #7421, opened 2025-07-25
chore: remove old custom web server api

keepview:
## Summary
- remove unstableMiddleware/mddleware/hook API
- c.set('forceCSR', '1') for fallback

## Checklist


packages/document/main-doc/docs/en/configure/app/server/ssr.mdx
```diff
@@ -62,13 +61,25 @@ By configuring `server.ssr.forceCSR` to `true` in the project, you can control t
...
-export const middleware = (ctx, next) => {
-  const { req, res } = ctx;
+```ts title="server/modern.server.ts"
+import {
+  defineServerConfig,
+  type MiddlewareHandler,
+} from '@modern-js/server-runtime';
+
+export const handler: MiddlewareHandler = async (c, next) => {
+  await next();
```
zllkjc: Should we exec `next()` after set Header ?

packages/server/core/src/plugins/render/render.ts
```diff
@@ -385,7 +388,8 @@ async function getRenderMode(
     const fallbackHeaderValue: string | null =
       (req.headers.get(fallbackHeader) as string) ||
       (nodeReq?.headers[fallbackHeader] as string);
-    if (forceCSR && (query.csr || fallbackHeaderValue)) {
+
+    if (forceCSR && (query.csr || fallbackHeaderValue || contextForceCSR)) {
```
zllkjc, line 392: Why do we need add `contextForceCSR`.

packages/document/main-doc/docs/en/configure/app/server/ssr.mdx
keepview: done

zllkjc (approved):

PR #7510, opened 2025-08-08
fix: solve some env problems

yimingjfe:
## Checklist


packages/solutions/app-tools/src/commands/build.ts
```diff
@@ -74,4 +111,5 @@ export const build = async (
   await appContext.builder.build({
     watch: options?.watch,
   });
+  await copyEnvFiles(appContext.appDirectory, appContext.distDirectory);
```
zllkjc: 这个 Copy 的位置可能会有坑。能不能在 AfterBuild 这种钩子里做。

yimingjfe: done

PR #7511, opened 2025-08-08
fix: correct route path construction in optimizeRoute function

yimingjfe:
## Summary

Fix the edge case in the pathless route. At present, when there is a pathless route, and no layout component under pathless route, and the direct sub-route under pathless route is dynamic route, this problem will be reproduced.

## Checklist

### DIFF
--- a/packages/runtime/plugin-runtime/src/router/cli/code/nestedRoutes.ts
+++ b/packages/runtime/plugin-runtime/src/router/cli/code/nestedRoutes.ts
@@ -81,9 +81,9 @@ export const optimizeRoute = (
     !routeTree.clientData
   ) {
     const newRoutes = children.map(child => {
-      const routePath = `${routeTree.path ? routeTree.path : ''}${
-        child.path ? `/${child.path}` : ''
-      }`;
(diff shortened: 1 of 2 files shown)


packages/runtime/plugin-runtime/src/router/cli/code/nestedRoutes.ts
```diff
@@ -81,9 +81,9 @@ export const optimizeRoute = (
     !routeTree.clientData
   ) {
     const newRoutes = children.map(child => {
-      const routePath = `${routeTree.path ? routeTree.path : ''}${
-        child.path ? `/${child.path}` : ''
-      }`;
+      const routePath = routeTree.path
```
clChenLiang, line 84: [routeTree.path, child.path]
.filter(Boolean)
.join(path.sep)

clChenLiang: [routeTree.path ?? '', child.path]
.filter(p => p !== undefined)
.join(path.sep)

clChenLiang (approved):

PR #7568, opened 2025-08-25
fix: solve the problem of react19 streaming rendering failure

yimingjfe:
## Checklist


packages/runtime/plugin-runtime/src/core/server/stream/createReadableStream.ts
```diff
@@ -100,6 +107,37 @@ export const createReadableStreamFromElement: CreateReadableStreamFromElement =
...
+                  ? Array.from(v7Active.entries())
+                  : v6Active
+                    ? v6Active instanceof Map
+                      ? Array.from(v6Active.entries())
+                      : Object.entries(v6Active)
+                    : [];
+
+              if (entries.length > 0) {
+                enqueueFromEntries(entries, config.nonce, (s: string) =>
+                  body.write(s),
```
zllkjc: I see that script write timing is controlled in `worker.ts`. Should we also control the timing of script push based on `shellChunkStatus` here?

packages/runtime/plugin-runtime/src/core/server/stream/createReadableStream.worker.ts
```diff
@@ -65,6 +67,41 @@ export const createReadableStreamFromElement: CreateReadableStreamFromElement =
...
+          const enqueueScript = (script: string) => {
+            if (shellChunkStatus === ShellChunkStatus.FINISH) {
+              controller.enqueue(encodeForWebStream(script));
+            } else {
+              pendingScripts.push(script);
+            }
+          };
+
+          const storageContext = storage.useContext?.();
+          const v7Active = storageContext?.activeDeferreds;
```
zllkjc: Why not using `isRouterV7` in this part.

packages/runtime/plugin-runtime/src/core/server/stream/deferredScript.ts
```diff
@@ -0,0 +1,90 @@
...
+}
+
+export function buildDeferredDataScript(
+  nonce: string | undefined,
+  args: unknown[],
+): string {
+  const payload = JSON.stringify(args);
+  const escaped = escapeAttr(payload);
+  const nonceAttr = nonce ? ` nonce=\"${nonce}\"` : '';
+  return `<script async${nonceAttr} data-fn-name=\"r\" data-script-src=\"modern-run-router-data-fn\" data-fn-args='${escaped}' suppressHydrationWarning>${runRouterDataFnStr}</script>`;
```
zllkjc: Why is there both `\"` and `'` here, It looks uncomfortable :)

packages/runtime/plugin-runtime/src/core/server/stream/createReadableStream.worker.ts
zllkjc: And I think it would be best to add some comments here, such as an introduction to the differences in data between React Router v6 and v7 here. This code is quite difficult to understand.

zllkjc (approved):

PR #7667, opened 2025-09-18
feat: use Rspack to compile document & support data fetch and ssgByEntries for ssg

yimingjfe:
## Summary

This PR mainly optimizes Document and SSG in the following ways:
1. Compile ⁠Document.tsx using Rspack, ensuring that ⁠Document.tsx and the project code are compiled by the same bundler. 	
2. Simplify SSG implementation logic by directly calling functions to trigger SSG rendering during build time, instead of sending requests. 	
3. Enhance SSG's data fetching and dynamic routing capabilities. 	
4. Add the `ssgByEntries` configuration for entries.
## Checklist


packages/runtime/plugin-runtime/src/document/Html.tsx
```diff
@@ -5,12 +5,22 @@ import { DocumentStructureContext } from './DocumentStructureContext';
 import { Head } from './Head';
 
 /**
- * get the directly son element
+ * get the directly son element by name
  */
-function findTargetChild(tag: string, children: ReactElement[]) {
+function findTargetChildByName(tag: string, children: ReactElement[]) {
   return children.find(item => getEleType(item) === tag);
```
clChenLiang, line 11: 从含义是，是不是命名 findTargetChildByTagName/Type 更合适

yimingjfe: 浏览器 dom 不是有 `getElementsByName` 这种 API 么？

PR #7839, opened 2025-11-06
feat(builder): add source.dedupe option

Kingwl:
## Summary
Add source.dedupe (which equals to rsbuild resolve.dedupe)

## Checklist


packages/cli/builder/src/shared/parseCommonConfig.ts
```diff
@@ -70,6 +70,7 @@ export async function parseCommonConfig(
     html: { outputStructure, appIcon, ...htmlConfig } = {},
     source: {
       alias,
+      dedupe,
```
chenjiahan, line 73: I prefer to add a new `resolve.dedupe` to align with Rsbuild 😄

Issue #7970, opened 2025-12-03
[Bug]: remote code execution vulnerability in bundled react-server-dom-webpack


jenseng:
### Version

```sh
System: *
Browsers: *
```

### Details

As of #6880, modern-js uses a vendored version of react-server-dom-webpack (19.0.0). Today (2025-12-03) Facebook disclosed a remote code execution vulnerability in this version. See:

- https://react.dev/blog/2025/12/03/critical-security-vulnerability-in-react-server-components
- https://www.cve.org/CVERecord?id=CVE-2025-55182

The React team has not (yet) published specifics about how the exploit works, but given that `@modern-js/render` uses react-server-dom-webpack to render RSCs, it seems highly likely that it is vulnerable

### Reproduce link

https://react.dev/blog/2025/12/03/critical-security-vulnerability-in-react-server-components

### Reproduce Steps

See https://react.dev/blog/2025/12/03/critical-security-vulnerability-in-react-server-components

sfarias051:
Hi everyone,

I've created a discussion thread about this critical vulnerability in the community Discord server, while keeping this GitHub issue focused on tracking the official fix implementation.

Maybe the Discord discussion could help coordinate efforts:
https://discord.com/channels/1373908733997875240/1446172984862707804/1446172984862707804

yimingjfe:
Hi @jenseng,

Thank you for reporting this, we'll fix this problem as soon as possible.

Modern.js's RSC feature is in the internal testing phase, so this vulnerability doesn't impact any repos temporarily.

yimingjfe:
This issue has been resolved, please upgrade to the latest version `2.69.3`

PR #7976, opened 2025-12-05
fix: upgrade react-server-dom-webpack to fix security vulnerability

yimingjfe:
## Summary

This PR upgrades `react-server-dom-webpack` to fix a critical security vulnerability (CVE-2025-55182) https://github.com/web-infra-dev/modern.js/issues/7970

## Checklist


micahgodbolt:
Is there an estimate for when this PR will be merged and published? We'd like to know if we can rely on this change happening soon or take other measure to patch the bundled dependency.

yimingjfe:
> Is there an estimate for when this PR will be merged and published? We'd like to know if we can rely on this change happening soon or take other measure to patch the bundled dependency.

Hi, we have released it, you can use version `2.69.3`, and this issue will not affect any Modern.js repositories，because only RSC projects will be affected by this issue, Modern.js has not release RSC feature.

PR #8223, opened 2026-01-22
perf: migrate from react-helmet to react-helmet-async for better SSR performance and async head management

GiveMe-A-Name:
## Summary

- Replace react-helmet with react-helmet-async to enable non-blocking, concurrent head tag rendering across SSR and hydration. 
- Improves SEO consistency and runtime performance under streaming/async scenarios, with corresponding updates to runtime integration and tests.

## Checklist

### DIFF
--- a/packages/runtime/plugin-runtime/src/exports/head.ts
+++ b/packages/runtime/plugin-runtime/src/exports/head.ts
@@ -1,4 +1,4 @@
-import { Helmet, HelmetProvider } from 'react-helmet-async';
+import { Helmet } from 'react-helmet-async';
 
 export default Helmet;
 


packages/runtime/plugin-runtime/src/exports/head.ts
```diff
@@ -1,5 +1,7 @@
-import head from 'react-helmet';
+import { Helmet, HelmetProvider } from 'react-helmet-async';
 
-export default head;
+export default Helmet;
 
-export * from 'react-helmet';
+export { Helmet, HelmetProvider };
+
+export * from 'react-helmet-async';
```
zllkjc: exported repeatedly?

packages/runtime/plugin-runtime/src/core/browser/hydrate.tsx
```diff
@@ -1,7 +1,7 @@
 import { loadableReady } from '@loadable/component';
-import { normalizePathname } from '@modern-js/runtime-utils/url';
 import type React from 'react';
 import type { Root } from 'react-dom/client';
+import { HelmetProvider } from '../../exports/head';
```
zllkjc: Is it okay to just import this directly from `react-helmet-async`...import from 'exports' seems strange

GiveMe-A-Name: Update~

packages/runtime/plugin-runtime/src/exports/head.ts
GiveMe-A-Name: Update!

PR #8270, opened 2026-01-29
docs: correct execution order of copy task

moonkop:
## Summary
While debugging copying complied css to another dist dir, I  found in fact [copy] is exexcuted before complie not after, document place copy task at last, for prevet misundestanding , I change the task order in doc 

## Checklist


Timeless0911 (approved):
Thanks

PR #8273, opened 2026-01-29
refactor: use node native fetch

sylingd:
## Summary

自 node 18 起，node 就已经内置 [Fetch API](https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API)，在 node 21 中被标记为 stable。目前 modern.js v2 要求的最低版本是 18+，可以移除原先使用的 node-fetch、isomorphic-fetch。

Since node 18, Node.js has had a built-in [Fetch API](https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API), which was marked as stable since with node 21. Currently, the minimum required version for modern.js v2 is 18+, and the previously used node-fetch and isomorphic-fetch can be removed.

## Checklist

### DIFF
--- a/.changeset/soft-views-sleep.md
+++ b/.changeset/soft-views-sleep.md
@@ -1,0 +1,5 @@
+---
+'@modern-js/create-request': minor
+---
+
+feat: use node native fetch


caohuilin:
@sylingd v2 分支也可以提交一份

PR #8310, opened 2026-02-04
fix(builder)!: align postcss object behavior with Rsbuild

9aoy:
## Summary

Align postcss object behavior with Rsbuild, make postcss object override instead of merge.

The postcss object handling in Modern.js is currently inconsistent with Rsbuild:
- Rsbuild applies override semantics
- Modern.js applies merge semantics

This divergence was accidentally introduced during the upgrade to Rsbuild 1.x. https://github.com/web-infra-dev/modern.js/commit/58f28689031429d5eedb332817388d287e4f11b4#diff-a86e848c67b89f2f9412c8fee114477c0ed805a5607500d47ddcebd5429595a5R67-R70
To maintain the previous behavior, several compatibility patches were later added in uni-builder, which increased complexity and maintenance cost.

This PR aligns Modern.js with Rsbuild by updating the postcss object behavior to use override semantics, matching Rsbuild’s configuration model.

## Related Links

https://rsbuild.rs/config/tools/postcss#object-type

## Checklist

### DIFF
--- a/packages/cli/builder/package.json
+++ b/packages/cli/builder/package.json
@@ -63,8 +63,7 @@
     "postcss-nesting": "12.1.5",
     "postcss-page-break": "3.0.4",
     "rspack-manifest-plugin": "5.2.1",
-    "ts-deepmerge": "7.0.3",
-    "jiti": "^2.6.1"
+    "ts-deepmerge": "7.0.3"
   },
   "devDependencies": {
     "@modern-js/rslib": "workspace:*",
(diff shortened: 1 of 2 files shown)


scripts/prebundle/src/constant.ts
```diff
@@ -127,32 +127,4 @@ export const TASKS: TaskConfig[] = [
...
   },
-  {
-    packageDir: 'cli/builder',
-    packageName: '@modern-js/builder',
-    dependencies: [
-      {
-        name: 'postcss-load-config',
-        externals: {
-          yaml: 'yaml',
-          jiti: 'jiti',
```
chenjiahan, line 138: We can remove the `jiti` dependency from `@modern-js/builder` package

chenjiahan (approved):

PR #8343, opened 2026-02-12
fix(builder): inlineScripts:false doesn't work

fzxen:
## Summary

inlineScripts always take their default value, even though I set it to false.

## Checklist

- [x] I have updated the documentation. - bugfix, no need
- [x] I have added tests to cover my changes. - my be latter


zllkjc:
Nice catch~

Please update the changeset to include both Chinese and English content, you may refer to previous commits.

caohuilin (approved):

PR #8433, opened 2026-03-10
fix: resolve React useId SSR hydration mismatch

GiveMe-A-Name:
## Summary

This PR fixes a hydration mismatch issue when using React's `useId` hook in SSR mode. The fix ensures consistent ID generation between server and client by introducing a new `useId` configuration option in the Modern.js runtime, allowing users to specify a unique identifier to prevent ID conflicts during hydration. Includes integration tests for SSR useId scenarios.

### DIFF
--- a/packages/runtime/plugin-runtime/src/core/constants.ts
+++ b/packages/runtime/plugin-runtime/src/core/constants.ts
@@ -6,7 +6,3 @@ export enum RenderLevel {
 export const SSR_DATA_JSON_ID = '__MODERN_SSR_DATA__';
 
 export const ROUTER_DATA_JSON_ID = '__MODERN_ROUTER_DATA__';
-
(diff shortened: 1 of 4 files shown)


packages/runtime/plugin-runtime/src/core/constants.ts
```diff
@@ -6,3 +6,7 @@ export enum RenderLevel {
 export const SSR_DATA_JSON_ID = '__MODERN_SSR_DATA__';
 
 export const ROUTER_DATA_JSON_ID = '__MODERN_ROUTER_DATA__';
+
+// React useId identifier prefix - re-export from @modern-js/utils for consistency
+// This ensures SSR and client hydration use the same ID prefix
+export { SSR_HYDRATION_ID_PREFIX } from '@modern-js/utils/universal/constants';
```
GiveMe-A-Name: TODO: remove it

Issue #8379, opened 2026-02-25
[Bug]: Netlify deploy failing due to missing netlify-handler.js


azinod:
### Version

```sh
System:
    OS: macOS 15.7.4
    CPU: (14) arm64 Apple M3 Max
    Memory: 301.03 MB / 36.00 GB
    Shell: 5.9 - /bin/zsh
  Browsers:
    Chrome: 145.0.7632.110
    Edge: 145.0.3800.70
    Firefox: 147.0.4
    Safari: 26.3
  npmPackages:
    @modern-js/app-tools: 3.0.2 => 3.0.2 
    @modern-js/runtime: 3.0.2 => 3.0.2 
    @modern-js/server-runtime: 3.0.2 => 3.0.2 
    @modern-js/tsconfig: 3.0.2 => 3.0.2
```

### Details

We started getting issues to deply to Netlify after upgrading to v3.0.2

When I run `MODERNJS_DEPLOY=netlify npx modern deploy`, the generated output files in `.netlify/functions` include a `netlify-handler.cjs`, however when I deploy to netlify the function initalization fails because it is looking for `netlify-handler.js` (regular JS file, not CJS)

I am guessing the issue might have to do with the line below that tries to import `netlify-handler.js` instead of `netlify-handler.cjs`
https://github.com/web-infra-dev/modern.js/blob/5269eac08597526590b729f8c5a29bd6ff05f534/packages/solutions/app-tools/src/plugins/deploy/platforms/netlify.ts#L141-L143

This is the error I get:
```
Feb 25, 02:55:57 PM: ERROR  Uncaught Exception 	{"errorType":"Runtime.ImportModuleError","errorMessage":"Error: Cannot find module './netlify-handler.js'\nRequire stack:\n- /var/task/index.js\n- /var/runtime/index.mjs","stack":["Runtime.ImportModuleError: Error: Cannot find module './netlify-handler.js'","Require stack:","- /var/task/index.js","- /var/runtime/index.mjs","    at _loadUserApp (file:///var/runtime/index.mjs:1192:17)","    at async UserFunction.js.module.exports.load (file:///var/runtime/index.mjs:1235:21)","    at async start (file:///var/runtime/index.mjs:1454:23)","    at async file:///var/runtime/index.mjs:1464:1"]}
```

Can you please check?

### Reproduce link

Don't have one yet

### Reproduce Steps

[...]

PR #8439, opened 2026-03-11
fix(netlify): support lambda event requests and cjs cache exports

azinod:
## Summary
This PR fixes two follow-up Netlify runtime regressions we hit while validating the Netlify deploy path after the bootstrap filename fix from #8379:

### 1. 
The first problem was that server-side CommonJS output could end up doing `require('@modern-js/runtime/cache')`, but the `./cache` subpath only exported an ESM target. In a real Netlify function this caused startup to fail with:

```text
Error [ERR_REQUIRE_ESM]: require() of ES Module /var/task/node_modules/@modern-js/runtime/dist/esm/cache/index.mjs from /var/task/server/lib/cache/setup.js not supported.
```

To fix that, this PR adds a CommonJS-compatible `node.require` export for `@modern-js/runtime/cache` and points the default export at the CJS build.

### 2. 
The second problem was in the Netlify request adapter. The generated Netlify handler was forwarding the Lambda-style Netlify event directly into `server.handle(...)`, but Hono expects a real Fetch `Request` with a `url`. In production this caused:

```text
TypeError: Cannot read properties of undefined (reading 'indexOf')
    at Hono.getPath (/var/task/node_modules/hono/dist/cjs/utils/url.js:103:21)
    at #dispatch (/var/task/node_modules/hono/dist/cjs/hono-base.js:298:23)
    at Hono.fetch (/var/task/node_modules/hono/dist/cjs/hono-base.js:347:26)
    at /var/task/node_modules/@modern-js/prod-server/dist/cjs/netlify.js:48:39
    at Runtime.handler (/var/task/netlify-handler.cjs:73:24)
```

To fix that, this PR teaches `@modern-js/prod-server/netlify` to:
- detect whether it already received a web `Request`
- otherwise convert the Netlify Lambda event into a Fetch `Request`
- pass that `Request` to `server.handle(...)`
- convert the returned `Response` back into a Netlify-compatible function response

Together these changes make the generated Netlify function artifact work in the downstream app we used for verification, both locally and in actual Netlify branch deploys.

## Related Links
- #8379
- Reproduced while testing downstream Netlify deployments generated by `MODERNJS_DEPLOY=netlify npx modern deploy`

## Checklist

## Validation
- `pnpm --filter @modern-js/runtime build`
- `pnpm --filter @modern-js/prod-server build`
- packed patched local Modern.js packages and consumed them from a downstream app
- verified `require.resolve('@modern-js/runtime/cache')` resolves to `dist/cjs/cache/index.js` in the generated Netlify function bundle
- verified the generated `@modern-js/prod-server/dist/cjs/netlify.js` contains the Netlify event -> `Request` adapter logic in the downstream app's `.netlify/functions` output
- validated the generated artifact path through Netlify branch deploys for `poc` and `develop`


clChenLiang:
@azinod let's close this mr, for has already finished in #8484

PR #8483, opened 2026-03-25
feat(server): support OPTIONS handler on ServerBase

### LINKED ISSUE #6148: [Feature]: Support OPTIONS method request for server on ModernJs

This helps us to be able handle the OPTIONS (method) request. When customizing server on ModernJs, I want to expose some APIs. Because these APIs are called on a different domain so browsers will send the OPTIONS request automatically. However, currently, ModernJs server can't catch the OPTIONS request so it throws the CORS error.

In the file `packages/server/core/src/serverBase.ts`.

I expect we can add the getter for `options`. Because it's duplicated with the name of property `options`, so I suggest that we'll use `serverOptions` for that property and have method `getOptions` to get its value

GiveMe-A-Name:
## Summary
- add `server.options(path, handler)` support in `ServerBase` so custom server code can handle `OPTIONS` requests
- avoid naming conflict by renaming internal config field to `serverOptions` and exposing `getOptions()`
- add unit and integration coverage for `OPTIONS` handling

## Testing
- `pnpm --filter @modern-js/server-core test -- packages/server/core/tests/plugins/favicon.test.ts`
- `pnpm --dir tests test:framework --runInBand --testPathPattern=integration/server-config/tests/index.test.ts`

Closes #6148

### DIFF
--- a/.changeset/fair-wasps-press.md
+++ b/.changeset/fair-wasps-press.md
@@ -1,0 +1,6 @@
+---
+'@modern-js/server-core': patch
+---
+
+feat: support OPTIONS handler on ServerBase
+feat: 为 serverBase 添加 options handler


zllkjc:
Thanks for the PR! A couple of things worth addressing:

**`getOptions()` is redundant alongside `public serverOptions`**

The PR renames `public options` to `public serverOptions` and adds a `getOptions()` accessor — but both are public, so there are now two ways to access the same data. If the intent of `getOptions()` is to provide a stable, encapsulated access path (which makes sense given the rename was needed to free up the `options` name), then `serverOptions` should be `private` or `protected`. Leaving it `public` while also exposing `getOptions()` creates an inconsistent API surface.

**Unit test belongs in the wrong file**

The new test for the OPTIONS handler is added to `favicon.test.ts`, which is semantically unrelated. It should live in a dedicated `serverBase.test.ts` (or similar) file that covers `ServerBase` behavior.

GiveMe-A-Name:
@zllkjc Addressed in 66cefb20c6.

- `serverOptions` is now `private` and `getOptions()` is kept as the single access path in `ServerBase`.
- Moved the OPTIONS handler test out of `favicon.test.ts` into `packages/server/core/tests/serverBase.test.ts`.
- Re-ran `pnpm --filter @modern-js/server-core test` locally (pass).

Please take another look.

zllkjc (approved):

PR #8500, opened 2026-04-01
fix(app-tools): restore dev server runtime stability for server config

GiveMe-A-Name:
## Summary
- prefer ts-node for dev server runtime loading when available
- preserve the Node 22 loader path for the existing server-config scenario
- fix bff-hono dev middleware and context behavior regression

## Verification
- cd tests && npm run test:framework --filter server-config
- cd tests && npm run test:framework --filter bff-hono

### DIFF
--- a/.changeset/empty-singers-push.md
+++ b/.changeset/empty-singers-push.md
@@ -1,0 +1,6 @@
+---
+'@modern-js/app-tools': patch
+---
+
+fix: restore dev server runtime stability for server config
+fix: 处理 dev server runtime 的稳定性


GiveMe-A-Name:
Updating the PR detail with the root cause and fix rationale here because `gh pr edit` is currently blocked by a GitHub Projects classic GraphQL error.

## Root cause
This regression started after `89fea12862db894d94b29bbfe94df2a97bd67a69`, which switched dev TypeScript runtime loading toward the Node 22 native loader path.

The regression only affects the dev server runtime path. `modern.server.ts` can still be loaded successfully, and its exported config shape is correct, but middleware/context behavior no longer matches the previous `ts-node` path in the affected scenario. In practice, this caused middleware defined in `modern.server.ts` to stop behaving consistently for `bff-hono` in dev, which then broke the Hono BFF context-dependent flow.

## Why this fix
Instead of rolling back the whole Node 22 loader change, this change applies a narrow fallback for the dev server runtime path: when `ts-node` is available, prefer it for this sensitive runtime-loading scenario.

This keeps the existing Node 22 loader behavior for the previously fixed `server-config` case, while restoring the stable middleware/context behavior required by `bff-hono` in dev.

## Verification
- `cd tests && npm run test:framework --filter server-config`
- `cd tests && npm run test:framework --filter bff-hono`

PR #8544, opened 2026-04-13
fix(create): add @rsbuild/core/types reference to modern-app-env.d.ts template

intellild:
## Summary

The following type reference is missing in the default modern-app-env.d.ts, which causes import.meta.xxx to result in a type error.

## Related Links

#8522 

## Checklist

### DIFF
--- a/packages/toolkit/create/template/src/modern-app-env.d.ts
+++ b/packages/toolkit/create/template/src/modern-app-env.d.ts
@@ -1,1 +1,2 @@
+/// <reference types="@rsbuild/core/types" />
 /// <reference types='@modern-js/app-tools/types' />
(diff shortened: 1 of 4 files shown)


packages/toolkit/create/template/src/modern-app-env.d.ts
```diff
@@ -1 +1,2 @@
+/// <reference types="@rsbuild/core/types" />
 /// <reference types='@modern-js/app-tools/types' />
```
caohuilin, line 1: Would it be better to add the `@rsbuild/core/types` reference in `@modern-js/app-tools/types` instead, so consumers only need to reference the app-tools types here?

intellild: Fixed in `53b785106a`: moved the `@rsbuild/core/types` reference into `@modern-js/app-tools/types`, restored the create template to only reference app-tools types, and added a focused regression test for that ownership.

caohuilin (approved):

PR #8595, opened 2026-05-07
feat(server): add compression support for production server

### LINKED ISSUE #6920: [Feature]: Gzip compression for production files

This feature request is a follow up on the discussion #6904 .

When using Modern.js server to serve output files directly, it would be beneficial to have a bundler.config option to set the compression of dist files to use `gzip`. This will enable developers to optimize the browser performance when receiving files dramatically. 
Although this is possible to be achieved via nginx configurations, etc., I am believe this is a great additional to the feature-stack of a modern full-stack-framework. Also, not all consumers rely on options like nginx to keep a clean footprint.

In the current version, Modern.js can use this compression method for devServer only:
```ts
export default defineConfig({
  devServer: {
    // Enable gzip compression in devServer config
    compression: true
  }
})
```

#### Bundler configuration
```ts
// modern.config.ts
export default defineConfig({
  devServer: {
    compression: true
  }
  server: {
    // Enable gzip compression in for production files as well
    compression: true
  }
})
```

#### Usage
The configuration should be used when starting the Modern.js server via `pnpm run start`, `pnpm run serve` and `pnpm run deploy`

jack-old-archive:
## Summary

Add gzip/br/deflate compression middleware for the production server via `server.compression` config option.

This allows developers to enable response compression without relying on a reverse proxy like nginx.

Usage:
```ts
// modern.config.ts

export default defineConfig({

  server: {

    compression: true, // gzip by default

    // or: compression: { encoding: 'br' }

  },

});


## Related Links

Closes #6920

## Checklist

<!--- Go over all the following points, and put an `x` in all the boxes that apply. -->

- [ x] I have added tests to cover my changes.

### DIFF
--- a/packages/runtime/plugin-runtime/src/core/server/stream/createReadableStream.ts
+++ b/packages/runtime/plugin-runtime/src/core/server/stream/createReadableStream.ts
@@ -130,7 +130,7 @@ export const createReadableStreamFromElement: CreateReadableStreamFromElement =
                       callback(e);
                     } else {
                       callback(
-                        new Error('Received unkown error when streaming'),
+                        new Error('Received unknown error when streaming'),
                       );
                     }
                   }
--- a/packages/runtime/plugin-runtime/src/router/runtime/constants.ts
+++ b/packages/runtime/plugin-runtime/src/router/runtime/constants.ts
@@ -34,7 +34,7 @@ export const resolveFnStr = `function r(e,r,o,A){A?_ROUTER_DATA.r[e][r].reject(A
 /**
    * update data for pre resolved promises
    * original function:
-   * function preResovledDeferredPromise(data, error) {
+   * function preResolvedDeferredPromise(data, error) {
     if(typeof error !== 'undefined'){
       return Promise.reject(new Error(error.message));
     }else{
--- a/packages/runtime/plugin-runtime/src/router/runtime/plugin.node.tsx
+++ b/packages/runtime/plugin-runtime/src/router/runtime/plugin.node.tsx
@@ -40,7 +40,7 @@ import {
 import type { RouterConfig } from './types';
 import { createRouteObjectsFromConfig, renderRoutes, urlJoin } from './utils';
 
-function createRemixReuqest(request: Request) {
+function createRemixRequest(request: Request) {
   const method = 'GET';
   const { headers } = request;
   const controller = new AbortController();
@@ -136,7 +136,7 @@ export const routerPlugin = (
 
         // We can't pass post request to query,due to post request would trigger react-router submit action.
         // But user maybe do not define action for page.
-        const remixRequest = createRemixReuqest(
+        const remixRequest = createRemixRequest(
           context.ssrContext!.request.raw,
         );
 
--- a/packages/server/core/src/plugins/compression.ts
+++ b/packages/server/core/src/plugins/compression.ts
@@ -1,0 +1,27 @@
+import { compress } from 'hono/compress';
+import type { ServerPlugin } from '../types';
+
+export const compressionPlugin = (): ServerPlugin => ({
+  name: '@modern-js/plugin-compression',
+  setup(api) {
+    api.onPrepare(() => {
+      const { middlewares, serverConfig } = api.getServerContext();
+      const compression = serverConfig?.server?.compression;
+
+      if (!compression) {
+        return;
+      }
+
+      const encoding =
+        typeof compression === 'object' && compression.encoding
+          ? compression.encoding
+          : 'gzip';
+
+      middlewares.push({
+        name: 'compression',
+        handler: compress({ encoding }),
+        order: 'pre',
+      });
+    });
+  },
+});
--- a/packages/server/core/src/plugins/index.ts
+++ b/packages/server/core/src/plugins/index.ts
@@ -14,3 +14,4 @@ export {
 } from './default';
 export { compatPlugin, handleSetupResult } from './compat';
 export { injectConfigMiddlewarePlugin } from './middlewares';
+export { compressionPlugin } from './compression';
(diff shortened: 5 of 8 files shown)

### CODE REVIEW

**Comment by keepview** [comment]
@Jack-sh1  Thanks for the PR — the motivation (compression without nginx/CDN) is valid. 👍

That said, I don't think this needs to live in the framework: Modern.js already exposes Hono middleware via the custom server (server/modern.server.ts), so users can enable hono/compress in a few lines:

```
// server/modern.server.ts
import { defineServerConfig, type MiddlewareHandler } from '@modern-js/server-runtime';
import { compress } from 'hono/compress';

export default defineServerConfig({
  middlewares: [
    { name: 'compress', order: 'pre', handler: compress() as MiddlewareHandler },
  ],
});
```

I verified this with modern serve: responses come back with content-encoding: gzip. This is basically what the plugin does internally, so server.compression is mostly a config alias over something already achievable.

A couple of issues with the current implementation, FWIW:
- brotli is broken — hono/compress uses the web-standard CompressionStream (gzip/deflate only); new CompressionStream('br') throws at runtime, so encoding: 'br' will crash.
- forcing encoding bypasses Accept-Encoding negotiation — Hono only negotiates when encoding is left unset.

Suggestion: document the recipe above instead of adding a new config/plugin. If you do want it built in, basing it on http-compression (already used by the dev server — supports brotli + a filter) would keep dev/prod consistent and make brotli actually work. Thanks again!

elio-Wang:
> @Jack-sh1 Thanks for the PR — the motivation (compression without nginx/CDN) is valid. 👍
> 
> That said, I don't think this needs to live in the framework: Modern.js already exposes Hono middleware via the custom server (server/modern.server.ts), so users can enable hono/compress in a few lines:
> 
> ```
> // server/modern.server.ts
> import { defineServerConfig, type MiddlewareHandler } from '@modern-js/server-runtime';
> import { compress } from 'hono/compress';
> 
> export default defineServerConfig({
>   middlewares: [
>     { name: 'compress', order: 'pre', handler: compress() as MiddlewareHandler },
>   ],
> });
> ```
> 
> I verified this with modern serve: responses come back with content-encoding: gzip. This is basically what the plugin does internally, so server.compression is mostly a config alias over something already achievable.
> 
> A couple of issues with the current implementation, FWIW:
> 
> * brotli is broken — hono/compress uses the web-standard CompressionStream (gzip/deflate only); new CompressionStream('br') throws at runtime, so encoding: 'br' will crash.
> * forcing encoding bypasses Accept-Encoding negotiation — Hono only negotiates when encoding is left unset.
> 
> Suggestion: document the recipe above instead of adding a new config/plugin. If you do want it built in, basing it on http-compression (already used by the dev server — supports brotli + a filter) would keep dev/prod consistent and make brotli actually work. Thanks again!

@keepview Thanks for the thorough review!

You're right about both issues — the brotli crash and the Accept-Encoding bypass are real bugs that need fixing. I'll address them.

That said, I'd still like to keep this as a built-in feature for a couple reasons:

1. **Not everyone uses a custom server** — many users just want compression out of the box without writing `server/modern.server.ts`. A one-line config (`server.compression: true`) is significantly lower friction.
2. **Dev/prod consistency** — the dev server already uses `http-compression` for compression. Having the same behavior in prod (via the same library) means users don't need to reason about two different compression paths.
3. **Common need** — issue #6920 has been open for a while, which suggests this is a recurring request.

I'll refactor this PR to:
- Replace `hono/compress` with `http-compression` (already used by the dev server)
- Fix brotli support (http-compression handles it natively)
- Properly negotiate `Accept-Encoding` headers
- Keep dev/prod behavior consistent

Would you be open to reviewing the updated implementation once it's ready?

keepview:
> > @Jack-sh1 Thanks for the PR — the motivation (compression without nginx/CDN) is valid. 👍
> > That said, I don't think this needs to live in the framework: Modern.js already exposes Hono middleware via the custom server (server/modern.server.ts), so users can enable hono/compress in a few lines:
> > ```
> > // server/modern.server.ts
> > import { defineServerConfig, type MiddlewareHandler } from '@modern-js/server-runtime';
> > import { compress } from 'hono/compress';
> > 
> > export default defineServerConfig({
> >   middlewares: [
> >     { name: 'compress', order: 'pre', handler: compress() as MiddlewareHandler },
> >   ],
> > });
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > I verified this with modern serve: responses come back with content-encoding: gzip. This is basically what the plugin does internally, so server.compression is mostly a config alias over something already achievable.
> > A couple of issues with the current implementation, FWIW:
> > 
> > * brotli is broken — hono/compress uses the web-standard CompressionStream (gzip/deflate only); new CompressionStream('br') throws at runtime, so encoding: 'br' will crash.
> > * forcing encoding bypasses Accept-Encoding negotiation — Hono only negotiates when encoding is left unset.
> > 
> > Suggestion: document the recipe above instead of adding a new config/plugin. If you do want it built in, basing it on http-compression (already used by the dev server — supports brotli + a filter) would keep dev/prod consistent and make brotli actually work. Thanks again!
> 
> @keepview Thanks for the thorough review!
> 
> You're right about both issues — the brotli crash and the Accept-Encoding bypass are real bugs that need fixing. I'll address them.
> 
> That said, I'd still like to keep this as a built-in feature for a couple reasons:
> 
> 1. **Not everyone uses a custom server** — many users just want compression out of the box without writing `server/modern.server.ts`. A one-line config (`server.compression: true`) is significantly lower friction.
> 2. **Dev/prod consistency** — the dev server already uses `http-compression` for compression. Having the same behavior in prod (via the same library) means users don't need to reason about two different compression paths.
> 3. **Common need** — issue [[Feature]: Gzip compression for production files #6920](https://github.com/web-infra-dev/modern.js/issues/6920) has been open for a while, which suggests this is a recurring request.
> 
> I'll refactor this PR to:
> 
> * Replace `hono/compress` with `http-compression` (already used by the dev server)
> * Fix brotli support (http-compression handles it natively)
> * Properly negotiate `Accept-Encoding` headers
> * Keep dev/prod behavior consistent
> 
> Would you be open to reviewing the updated implementation once it's ready?

sure~

PR #8621, opened 2026-05-19
feat: include server directory in tsconfig.json from @modernjs/create template

s-r-x:
## Summary
`modern/modern.server.ts` is ignored if the folder isn't included in `tsconfig.json`

## Checklist


zllkjc:
Thanks for your contribution. Could you please update the ZH documentation as well?

s-r-x:
> Thanks for your contribution. Could you please update the ZH documentation as well?

done. my mandarin is a bit rusty, so i used ai for that

caohuilin (approved):

PR #8658, opened 2026-06-04
fix(runtime): don't dedupe SSR route stylesheets against non-stylesheet <link> tags

deepcoldy:
## Summary

In **string-mode SSR**, a route's CSS can be silently dropped (downloaded but never applied) when the same CSS file is also referenced by a non-stylesheet `<link>` in the HTML template — most commonly a `<link rel="prefetch">` emitted by `performance.prefetch`. The first screen then renders **unstyled**.

## Root cause

`LoadableCollector.emitStyleAssets` (`packages/runtime/plugin-runtime/src/core/server/string/loadable.ts`) builds the set of "already present" links to dedupe against using a regex that matches **every** `<link href>`:

```ts
const linkRegExp = /<link .*?href="([^"]+)".*?>/g;
// ...collects href of EVERY <link>, including rel="prefetch" / rel="preload" / rel="icon"
```

It then skips injecting any collected route stylesheet whose url is already in that set. When `performance.prefetch` is enabled it adds `<link rel="prefetch" href=".../page.css">` to the template, so the real `<link rel="stylesheet" href=".../page.css">` is considered "already injected" and skipped. The CSS is prefetched (fetched into cache) but never applied.

Entry/initial CSS is unaffected because it is injected by the HTML plugin independently. Scripts are unaffected because `emitScriptAssets` matches `<script src>`, and prefetched JS is a `<link rel="prefetch">` (not a `<script>`), so it never collides.

## Reproduction

- App with SSR on a route whose CSS is a code-split (async) chunk, **string mode** (`renderToString`, the default).
- `performance: { prefetch: { type: 'async-chunks' } }` (or any prefetch that emits `<link rel="prefetch">` for css).
- Result: the route's `page.css` appears in the SSR head only as `<link rel="prefetch">`, never as `<link rel="stylesheet">`; the page renders without its styles.

## Fix

Only dedupe against existing `<link rel="stylesheet">` tags. `rel="prefetch"`/`rel="preload"`/`rel="icon"` links may point at the same file but do not apply styles, so they must not suppress the real stylesheet injection.

## Verification

Reproduced and verified end-to-end on a real Modern.js/EdenX app (string-mode SSR + `performance.prefetch: { type: 'async-chunks' }`):

- **Before:** the route CSS (`home_page.css`) was present only as `<link rel="prefetch">`; page unstyled.
- **After** (this dedupe fix / dropping the css prefetch): the SSR head contains `<link rel="stylesheet" href=".../home_page.css">` and the page is styled.

With this fix, `performance.prefetch` and string-mode SSR work together without needing a per-app `exclude: [/\.css$/]` workaround.

## Notes

`emitStyleAssets` is a private method that depends on `@loadable/server`'s `ChunkExtractor`, so it isn't covered by a focused unit test today. Happy to add one (e.g. by extracting the "collect existing stylesheet hrefs" step into a small pure helper) if you'd prefer it in this PR.


deepcoldy:
Superseded by #8699, which fixes string-mode SSR using the shared `hasStylesheetLink` helper introduced in #8694 (rather than a separate inline regex). Closing in favor of that.

PR #8692, opened 2026-06-15
fix(plugin-runtime): fix document CLI compatibility with ESM lib format

imhele:
### Problem

The document plugin's CLI helpers (`requireFromString` and `generateEntryCode`) relied on CJS-only Node.js APIs that are unavailable or incorrect when the package is built and consumed as an ES module:

1. **`module.parent` in `requireFromString`**: In ESM context, the CommonJS `module` global does not exist, so passing `module.parent` as the parent to `new Module.Module()` throws a `ReferenceError` at runtime.

2. **`require.resolve('../')` in `generateEntryCode`**: `createRequire().resolve('../')` cannot resolve to `index` file in an ESM module context.

### Solution

Both helpers are now guarded by a `process.env.MODERN_LIB_FORMAT === 'esm'` check so they use the appropriate API depending on the module format:

- **`requireFromString`**: passes `undefined` as the parent module when in ESM mode, which is a safe no-op for `Module.Module` and avoids the CJS `module.parent` reference.
- **`generateEntryCode`**: uses `import.meta.resolve('../index.mjs')` to resolve the runtime API path in ESM mode (the `.mjs` entry is already the correct target), falling back to `require.resolve('../')` in CJS mode.

### Affected package

- `@modern-js/runtime` (patch)

## Checklist

### DIFF
--- a/packages/runtime/plugin-runtime/src/document/cli/index.ts
+++ b/packages/runtime/plugin-runtime/src/document/cli/index.ts
@@ -181,7 +181,12 @@ const processCommentPlaceholders = (html: string): string => {
 
 // load CommonJS module from code string (evaluated in Node), returns exports
 const requireFromString = (code: string, filename: string) => {
-  const m = new Module.Module(filename, module.parent as Module);
+  const m = new Module.Module(
+    filename,
+    process.env.MODERN_LIB_FORMAT === 'esm'
+      ? undefined
+      : (module.parent as Module),
+  );
   m.filename = filename;
   // set proper resolution paths for nested requires
   // eslint-disable-next-line @typescript-eslint/ban-ts-comment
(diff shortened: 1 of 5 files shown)


zllkjc:
Thanks for contribute~ Can you add a Chinese version changeset ?

imhele:
> Thanks for contribute~ Can you add a Chinese version changeset ?

Hi, added another ESM bug fix. Verified in my project by pnpm-patches, works well.

packages/runtime/plugin-runtime/src/document/cli/index.ts
```diff
@@ -269,10 +274,14 @@ const applyExternalsPlugin = (child: Compiler, compiler: Compiler) => {
...
 
 const generateEntryCode = (docPath: string, _entryName: string): string => {
-  const runtimeAPI = require.resolve('../');
-  const esmRuntimeAPI = runtimeAPI
-    .replace(`cjs`, `esm`)
-    .replace(/.js$/, '.mjs');
+  const runtimeAPI =
+    process.env.MODERN_LIB_FORMAT === 'esm'
+      ? // @ts-ignore
+        import.meta.resolve('../index.mjs')
```
zllkjc: 这里会解析到 `dist/esm-node/document/index.mjs`，而用户侧 `@modern-js/runtime/document` 指向 `dist/esm/document/index.mjs`，导致 Provider 和 Consumer 使用不同的 `DocumentContext` 实例、无法读取上下文值。可以直接暴力点，替换成 dist/esm 下的那个

---

This resolves to `dist/esm-node/document/index.mjs`, while user code imports `dist/esm/document/index.mjs`, so the Provider and Consumer use different `DocumentContext` instances and the context value is lost.

zllkjc (approved):

PR #8735, opened 2026-07-01
fix(bff): make cross-project client declarations portable

### LINKED ISSUE #8739: [Bug]: BFF crossProject client .d.ts unresolvable for consumers (leaked path aliases + depth-shifted relative imports)
### Version

```sh
System: macOS 15 (darwin 24.6)
npmPackages: @modern-js/* 3.5.x (also reproduces on current main)
```

### Details

With `bff.crossProject: true` and `declaration: true`, the generated `<package>/api/*` client declarations cannot be resolved by consuming projects. Two independent defects compound:

1. **tsconfig path aliases leak into every emitted `.d.ts`.** `@modern-js/server-utils` applies its tsconfig-paths transformer only as a `before` custom transformer, and TypeScript runs those solely on the JS emit — declaration emit keeps aliases as authored (see microsoft/TypeScript#30952). The emitted JS gets `require("../shared/index")` while the sibling `.d.ts` keeps `from '@shared/index'`, which no consumer can resolve.
2. **The cross-project client copy breaks relative specifiers.** `clientGenerator` copies each handler's declaration from `dist/<lambda>/…` into `dist/client/…` verbatim. The copy lands one directory shallower than its origin, so relative specifiers TypeScript emitted for the origin (including inline `import("…")` types) no longer resolve from the copy.

A consumer importing the client types fails with errors like:

```
error TS2307: Cannot find module '@shared/index' or its corresponding type declarations.
```

### Reproduce link

Reproducible with the in-repo fixture `tests/integration/bff-corss-project` (see reproduce steps).

### Reproduce Steps

1. In `bff-api-app` (has `crossProject: true`, tsconfig `paths`, `declaration: true`), give a handler an aliased type signature, e.g. in `api/lambda/index.ts`:
   ```ts
   import type { ApiMessage } from '@shared/index';
   export default async (): Promise<ApiMessage> => ({ message: 'hi' });
   ```
2. Run `modern build` in `bff-api-app`.
3. Inspect `dist-1/client/index.d.ts` — it contains `from '@shared/index'` (unresolvable outside the project). Handlers whose declarations carry relative/inline `import("…")` specifiers additionally point one directory too deep after the copy.

omarshibli:
## Summary

With `bff.crossProject: true` and `declaration: true`, consuming projects cannot resolve the generated `<package>/api/*` client types, failing with errors like:

```
error TS2307: Cannot find module '@shared/index' or its corresponding type declarations.
```

Two independent defects compound here:

1. **Path aliases leak into every emitted declaration.** `compileByTs` applies the tsconfig-paths transformer only as a `before` custom transformer, and TypeScript runs those solely on the JS emit — declaration emit keeps aliases exactly as authored (microsoft/TypeScript#30952). So the emitted JS gets `require("../shared/index")` while the sibling `.d.ts` keeps `from '@shared/index'`.
2. **The cross-project client copy breaks relative specifiers.** `clientGenerator` copies each handler's declaration from `dist/<lambda>/…` into `dist/client/…` verbatim. The copy lands one directory shallower than its origin, so the relative specifiers TypeScript emitted for the origin (including inline `import("…")` types) no longer resolve from the copy.

### Fix

Each defect is fixed in the layer that creates it:

- **`@modern-js/server-utils`** — the existing tsconfig-paths transformer now also runs on declaration emit (`afterDeclarations`), reusing the same path matcher. The declaration variant additionally handles inline `import("…")` types (`ImportTypeNode`), reads specifiers via `.text` (declaration nodes are synthesized without source positions, so `getText()` is unavailable), and keeps specifiers extensionless, matching tsc's own declaration output.
- **`@modern-js/plugin-bff`** — the client generator re-bases relative specifiers onto the copy's location at copy time, where the origin and destination of each file are both known.

Because both fixes live where dev and build converge, dev output is fixed as well, and no post-build processing step is needed.

Real output from the `bff-corss-project` integration fixture, before → after:

```ts
// dist/api/lambda/index.d.ts (origin)
import type { ApiMessage } from '@shared/index';        // before
import type { ApiMessage } from "../../shared/index";   // after

// dist/client/index.d.ts (generated client, one directory shallower)
import type { ApiMessage } from '@shared/index';        // before (verbatim copy)
import type { ApiMessage } from "../shared/index";      // after (re-based)
```

Bare package specifiers (e.g. `import("@modern-js/plugin-bff/server")`) are left untouched.

### Behavior notes

All `server-utils` `compile` consumers (the BFF api compile and app-tools' server compile) now get alias-free declarations when `declaration` is enabled. The previously emitted aliased specifiers were unresolvable outside the project, and the JS emit already received the identical rewrite, so this is strictly a fix rather than a behavior change consumers could have relied on.

### Test plan

- `packages/server/utils/tests/ts.test.ts` — new declaration-emit case: an aliased named import and an inline `import("…")` type are rewritten to relative specifiers; existing JS-emit cases unchanged.
- `packages/cli/plugin-bff/tests/clientGenerator.test.ts` — unit tests for specifier re-basing (`from` / `import(…)` / `require(…)` forms; bare specifiers untouched; no-op when origin and target match).
- `tests/integration/bff-corss-project` — `bff-api-app` now has an aliased type in a handler signature; a new prod-suite test walks every generated `dist-1/client/**/*.d.ts`, asserts no path alias leaks, and asserts every relative specifier resolves inside the dist tree.
- Verified locally: `server-utils` and `plugin-bff` unit suites plus the full `bff client-app in prod` integration describe are green.

## Related Links

- Fixes #8739

## Checklist

- [x] I have updated the documentation. (N/A — bug fix, no API change)

### DIFF
--- a/.changeset/bff-portable-client-types.md
+++ b/.changeset/bff-portable-client-types.md
@@ -1,0 +1,7 @@
+---
+'@modern-js/plugin-bff': patch
+---
+
+fix(plugin-bff): re-base relative specifiers when copying cross-project client declarations
+
+The cross-project client generator copied each handler's declaration into `dist/client` verbatim, one directory shallower than its origin, which broke the relative specifiers TypeScript emits. Specifiers are now re-based onto the copy's location so the generated `<package>/api/*` client types resolve in consuming projects.
--- a/.changeset/server-utils-declaration-aliases.md
+++ b/.changeset/server-utils-declaration-aliases.md
@@ -1,0 +1,7 @@
+---
+'@modern-js/server-utils': patch
+---
+
+fix(server-utils): resolve tsconfig path aliases in emitted declaration files
+
+The tsconfig-paths transformer only ran on the JS emit, so path aliases (unresolvable outside the project) leaked into `.d.ts` output when `declaration` is enabled. The same rewrite now also runs on declaration emit, including inline `import("...")` types.
--- a/packages/cli/plugin-bff/src/utils/clientGenerator.ts
+++ b/packages/cli/plugin-bff/src/utils/clientGenerator.ts
@@ -212,9 +212,40 @@ async function setPackage(
   }
 }
 
-export async function copyFiles(from: string, to: string) {
+const RELATIVE_SPECIFIER =
+  /(\bfrom\s*|\bimport\(\s*|\brequire\(\s*)(['"])(\.\.?\/[^'"]*)\2/g;
+
+// The generated client mirrors the handler's declaration but lives one
+// directory shallower (`dist/client/*` vs `dist/<lambda>/*`), so the relative
+// specifiers TypeScript emitted for the origin file must be re-based to keep
+// resolving from the copy's location.
+export function rebaseDeclarationSpecifiers(
+  source: string,
+  fromDir: string,
+  toDir: string,
+) {
+  return source.replace(
+    RELATIVE_SPECIFIER,
+    (_match, lead, quote, specifier) => {
+      const target = path.resolve(fromDir, specifier);
+      const relative = toPosixPath(path.relative(toDir, target));
+      const rebased = relative.startsWith('.') ? relative : `./${relative}`;
+      return `${lead}${quote}${rebased}${quote}`;
+    },
+  );
+}
+
+async function copyDeclarationFile(from: string, to: string) {
   if (await fs.pathExists(from)) {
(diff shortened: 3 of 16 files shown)


omarshibli:
Hey @keepview, would appreciate your input 🙏

omarshibli:
fixed here https://github.com/web-infra-dev/modern.js/pull/8797

PR #8752, opened 2026-07-07
fix(builder): respect output.module with ESM output in the service-worker environment

BleedingDev:
## Problem
The service-worker environment always rewrites the bundler output library type to `commonjs2`.
When a user enables `output.module`, the environment still ends up with a CJS library wrapper, so ESM worker output is not preserved.

## Fix
Detect service-worker bundler chains that already have module output enabled.
Use `type: 'module'` for that case and keep `commonjs2` for existing non-module service-worker output.

## Test
Added a builder unit test for the service-worker module-output branch.
Ran `PUPPETEER_SKIP_DOWNLOAD=1 pnpm install`, `pnpm --filter @modern-js/builder build`, and `pnpm --filter @modern-js/builder test`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


zllkjc (approved):

zllkjc:
Thanks for contribution! Approval~

PR #8753, opened 2026-07-07
feat(runtime): add viewport prefetch behavior to PrefetchLink

BleedingDev:
## Problem

`PrefetchLink` supports `intent`, `render`, and `none`, but cannot defer prefetching until a link is near the viewport. That makes long lists of links choose between prefetching too early or waiting for user interaction. Route chunk preloads were also tied to render-time work, which made viewport-triggered prefetch harder to model cleanly.

## Fix

Add `prefetch="viewport"` for `Link` and `NavLink` using `IntersectionObserver` with a viewport margin. Route chunks now preload from the prefetch link effect while existing data prefetch links still use the SSR data path.

## Test

Added a router runtime unit test for viewport-triggered prefetch. Ran install, scoped runtime build, and scoped runtime tests.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


yimingjfe:
Thanks for the PR. After looking at this again, I’m not sure we need a built-in `viewport` mode.

Apps that need viewport-based prefetching can use their existing visibility observer and switch the current prop between `none` and `render`:

```tsx
<Link prefetch={isNearViewport ? 'render' : 'none'} />
```

This also lets apps share observers and choose their own margin, threshold, virtual-list integration, or network conditions.

Unless there’s a use case that cannot be covered by controlling the existing prop, I’d prefer to keep the API small and not add `prefetch="viewport"` for now.

PR #8783, opened 2026-07-28
feat(app-tools): support programmatic dev/build lifecycle

GiveMe-A-Name:
Use the app context command to initialize builders and resolve occupied ports for programmatic dev and start calls.

Export closeServer from the package root, await Node close callbacks, and propagate close errors.

## Checklist


deepcoldy (approved):
LGTM. Reviewed the programmatic dev/build/deploy lifecycle changes; second review by an independent reviewer confirmed no blockers.

Verified:
- Dist cleanup restored to `onPrepare` (runs before analyze's `generateEntryCode`), so `dist/nestedRoutes.json` is preserved — confirmed via a real `modern build` on the routes integration fixture. Command detection now recognizes both the CLI argv command and the programmatic `appContext.command`.
- Programmatic `deploy` is intentionally excluded from the context fallback so `deploy({ skipBuild: true })` preserves existing dist; programmatic-deploy dist cleanup is an accepted known limitation. CLI `deploy --skip-build` keeps its original semantics.
- Dev restart behavior is equivalent to the previous implementation.
- `pnpm --filter @modern-js/app-tools build` + `tsc --noEmit` pass; app-tools tests 110/110 pass; Biome clean; CI all green.

Non-blocking nit: the PR description still mentions closeServer "await Node close callbacks, and propagate close errors", which no longer matches the reverted fire-and-forget implementation — worth removing.

PR #8797, opened 2026-08-06
fix(bff): make published crossProject client declarations portable

### LINKED ISSUE #8739: [Bug]: BFF crossProject client .d.ts unresolvable for consumers (leaked path aliases + depth-shifted relative imports)

keepview:
## Problem

`bff.crossProject` client `.d.ts` resolve locally but break once **published** (fixes #8739, supersedes #8735): consumers hit `TS2307` or the type degrades to `any`. Three causes:

1. **Alias leak** — tsc keeps tsconfig `paths` verbatim in `.d.ts` ([TS#30952](https://github.com/microsoft/TypeScript/issues/30952)), so `@shared/*` reaches consumers unresolved.
2. **Broken relative specifiers** — the client `.d.ts` is copied one directory shallower than tsc emitted it, mispointing every relative import.
3. **Missing files** — `dist/shared` / `dist/api` are never added to `package.json#files`, so they're absent from the tarball.

## Changes

- **server-utils**: rewrite tsconfig aliases on declaration emit via a new `afterDeclarations` transformer (Import/Export/ImportType/ImportEquals), keeping specifiers relative and extensionless. No public API change.
- **plugin-bff**: the client re-exports the handler declaration in place via a facade instead of copying it; publish every `.d.ts` (`<distPath>/**/*.d.ts`); and drive `readDirectoryFiles` from `ApiRouter.getApiFiles()` so stray files never reach `generateClient`.

## Testing

Unit (alias rewrite, facade, file filtering) + an out-of-repo tarball consumer that `tsc`s the published client with `skipLibCheck: false` and asserts no leak and no `any` degradation.

Co-Authored-By: Riff

### DIFF
--- a/.changeset/bff-crossproject-portable-declarations.md
+++ b/.changeset/bff-crossproject-portable-declarations.md
@@ -1,0 +1,8 @@
+---
+'@modern-js/server-utils': patch
+'@modern-js/plugin-bff': patch
+---
+
+fix: make published crossProject BFF client declarations resolvable in consumers
+
+fix: 修复发布后的 crossProject BFF 客户端声明在消费方项目中无法解析类型的问题
--- a/packages/cli/plugin-bff/src/cli.ts
+++ b/packages/cli/plugin-bff/src/cli.ts
@@ -78,7 +78,7 @@ export const bffPlugin = (): CliPlugin<AppTools> => ({
     };
 
     const generator = async () => {
-      const { appDirectory, apiDirectory, lambdaDirectory, port } =
+      const { appDirectory, apiDirectory, lambdaDirectory, port, moduleType } =
         api.getAppContext();
 
       const modernConfig = api.getNormalizedConfig();
@@ -122,6 +122,7 @@ export const bffPlugin = (): CliPlugin<AppTools> => ({
         httpMethodDecider,
         relativeDistPath,
         relativeApiPath,
+        moduleType,
         apiFiles: apiRouter.getApiFiles(),
       });
       await runtimeGenerator({
--- a/packages/cli/plugin-bff/src/cli.ts
+++ b/packages/cli/plugin-bff/src/cli.ts
@@ -122,6 +122,7 @@ export const bffPlugin = (): CliPlugin<AppTools> => ({
         httpMethodDecider,
         relativeDistPath,
         relativeApiPath,
+        apiFiles: apiRouter.getApiFiles(),
       });
       await runtimeGenerator({
         runtime,
--- a/packages/cli/plugin-bff/src/utils/clientGenerator.ts
+++ b/packages/cli/plugin-bff/src/utils/clientGenerator.ts
@@ -15,6 +15,12 @@ export type APILoaderOptions = {
   relativeDistPath: string;
   relativeApiPath: string;
   /**
+   * `'module'` for native ESM output (`package.json#type: "module"`); the
+   * generated facade then re-exports with an explicit `.js` extension so
+   * `node16`/`nodenext` consumers resolve it.
+   */
+  moduleType?: 'module' | 'commonjs';
+  /**
    * Absolute paths of the valid API files, resolved by ApiRouter with the same
    * `API_FILE_RULES` the runtime router uses. Passing them in keeps the client
    * generator and the router in agreement about what an API module is, so stray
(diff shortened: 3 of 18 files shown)


zllkjc (approved):

SoonIter:
<agent: daily-job rstack ecosystem-ci>

### Ecosystem CI regression: `rspack/modernjs`

**Attribution:** Confirmed downstream source for the four `bff-corss-project` failures

**Impact:** Modern.js framework tests now fail the cross-project BFF dev assertions on both Linux and Windows; the client receives fallback HTML or stale placeholder content instead of the BFF responses.

#### Why this PR

This PR added `types-portability.test.ts`, which calls `modernBuild(apiAppDir)` on the same `bff-api-app` fixture while `index.test.ts` runs its dev servers from that directory in a parallel test file. The post-merge Linux log shows the portability build removing `dist-1/client/index.js` and other generated files and triggering a server-runtime reload; the sibling dev assertions then fail with missing `.hello`, HTML parsed as JSON, and stale `bff-client` / `bff-express` text. The same four failures occur on the merge commit's Linux and Windows integration jobs and in ecosystem CI, while the previous ecosystem run used the pre-merge Modern.js main and passed.

#### Suggested next step

Modern.js tests should build and pack an isolated copy of `bff-api-app` in `types-portability.test.ts`, so it cannot mutate the live fixture used by `index.test.ts`. A fallback is to serialize these two files, though fixture isolation is safer. Verify the focused `integration/bff-corss-project` tests repeatedly on Linux and Windows, then rerun `pnpm test --stack rspack modernjs`.

<details>
<summary>Evidence and attribution checks</summary>

- Current run: https://github.com/rstackjs/rstack-ecosystem-ci/actions/runs/31670996525
- Tested upstream: `rspack` at `be2daae9a14bf0ad7de13f6fd24964071c120f4d`
- Previous good: https://github.com/rstackjs/rstack-ecosystem-ci/actions/runs/31662414674/job/94329850400 (pre-merge Modern.js main)
- First bad / source commit: `8edf91adb1819dc835a300255417d966c7d11747` from this PR
- Source CI confirmation: Linux https://github.com/web-infra-dev/modern.js/actions/runs/31667295291/job/94344487506 and Windows https://github.com/web-infra-dev/modern.js/actions/runs/31667294976/job/94344486255
- Failure signature: four `bff-corss-project` dev assertions fail after the new portability build mutates `dist-1`; 4 failed / 370 passed on the source Linux job
- Flaky/pre-existing check: the same BFF signature is absent from the previous eco-ci green control and appears deterministically on both source-commit platforms
- Config check: not applicable; this is cross-file fixture mutation in the Modern.js integration suite
- Alternatives ruled out: Rspack #15175 only changes Cargo dependencies for 32-bit cache compilation and does not touch the BFF/runtime path. The extra eco-ci-only `pure-esm-project/deploy` failure is not part of this attribution.

</details>

PR #8838, opened 2026-08-28
fix(runtime): serialize deferred resolver scripts safely

GiveMe-A-Name:
## Summary

- Serialize deferred resolver scripts through a shared HTML-safe output coordinator.
- Keep `ShellChunkStatus.FINISH` as the pre-shell release gate for Node and Edge.
- Wait for React continuation-protocol completion and registered resolver promises before closing.

## Validation

- `pnpm --filter @modern-js/runtime build`
- `pnpm --filter @modern-js/runtime test`
- `pnpm --dir tests exec rstest integration/ssr/tests/streaming.test.ts`

## Review

- Independent review completed; added coverage for HTML, SVG, MathML, and table continuation wrappers.

### DIFF
--- a/packages/runtime/plugin-runtime/src/core/server/stream/deferredScriptOutputCoordinator.ts
+++ b/packages/runtime/plugin-runtime/src/core/server/stream/deferredScriptOutputCoordinator.ts
@@ -81,9 +81,33 @@ function getSuffix(value: string, pattern: string): string {
 }
 
 /**
- * Serializes deferred resolver scripts with the React output stream without
- * rewriting React HTML. Shell completion makes a resolver eligible to send;
- * this coordinator waits for a root-level HTML boundary before emitting it.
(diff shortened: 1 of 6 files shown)


packages/runtime/plugin-runtime/src/core/server/stream/deferredScriptOutputCoordinator.ts
```diff
@@ -0,0 +1,547 @@
...
+
+    while (this.pendingResolvers.length > 0) {
+      const resolver = this.pendingResolvers.shift();
+      if (resolver) {
+        this.emit(resolver);
+      }
+    }
+  }
+
+  private consume(char: string): void {
```
GiveMe-A-Name, line 240: Maybe we can use react hunk size or anther way to match.

zllkjc (approved):

PR #8873, opened 2026-09-15
fix(app-tools): only externalize server-runtime when the app can resolve it

keepview:
## Problem

`@modern-js/plugin-server-runtime` externalizes `@modern-js/server-runtime` unconditionally:

```ts
api.config(() => ({
  output: {
    externals: [{ '@modern-js/server-runtime': '@modern-js/server-runtime' }],
  },
}));
```

The SSR bundle therefore keeps a `require('@modern-js/server-runtime')`. Frameworks built on Modern.js re-export that package under their own name, and their apps depend on the wrapper, so the bare specifier is only a transitive dependency and does not resolve from the app root under pnpm. Loading the bundle throws and `loadBundle` swallows it, so rendering silently degrades to CSR with a single log line:

```
bundle failed, error = Error: Cannot find module '@modern-js/server-runtime'
```

`@modern-js/app-tools` does not depend on `@modern-js/server-runtime` itself, so the external has always relied on the app declaring it.

## Why the external is no longer required

It was added in #7236 to keep the loader and the custom server on one `AsyncLocalStorage`, back when the storage was module-scoped. #8776 (shipped in 3.8.0) moved that storage to a process-global `globalThis[Symbol.for(...)]` registry, so duplicated copies already share the request context. The external now only carries its downside.

Deleting the plugin outright was considered and rejected: `@modern-js/server-runtime` re-exports from the `@modern-js/server-core` package root, which would pull the whole server-core graph (including module-level state such as the SSR memory cache in `plugins/render/ssrCache.ts`) into every loader bundle. Apps that can resolve the package are better off keeping it external.

## Fix

Externalize only when the package resolves from the app directory; otherwise leave it bundled.

| App | Resolves | Behavior | Change |
| --- | --- | --- | --- |
| Declares `@modern-js/server-runtime` | yes | external, as before | none |
| Depends on a wrapper package instead | no | bundled copy | error to working |

## Verification

- All integration fixtures using `useHonoContext` declare the dependency, so they keep the existing path: `bff-hono`, `bff-corss-project/bff-api-app`, `bff-corss-project/bff-client-app`.
- `integration/bff-hono` and `integration/bff-corss-project` pass, including `support useContext` in dev and prod.
- New unit test covers both branches.
- `tsc --noEmit` and `pnpm run build` pass.

Co-Authored-By: Riff


keepview:
Superseded by #8875, which removes the plugin outright. Closing in favour of that.

PR #8874, opened 2026-09-15
fix(builder): keep server chunk ids numeric to avoid filename collisions

keepview:
## Problem

Rsbuild 2.2.6 re-enabled compact hashed chunk ids in production (web-infra-dev/rsbuild#8478). Server output places entry and async chunks in one flat directory and names them `[name].js`:

```ts
// packages/cli/builder/src/plugins/environmentDefaults.ts
compatConfig.environments.server = {
  output: {
    emitAssets: false,
    distPath: { root: 'dist/bundles', js: '', css: '', jsAsync: '', cssAsync: '' },
  },
};
```

An unnamed async chunk falls back to its chunk id for `[name]`, so a short hashed id can collide with an entry name. A two-entry SSR app whose entries are `a` and `b` now fails the server build:

```
error   build failed (server)
× Conflict: Multiple assets emit different content to the same filename a.js

SSR Error - SSR render fallback, error = Error: Can't find renderBundle a
error   Error: Rspack build failed.
```

The render bundle never gets written, so rendering silently falls back to CSR.

## Isolation

Reproduced on a two-entry SSG/SSR fixture, then bisected:

| Setup | Result |
| --- | --- |
| Rsbuild 2.2.4 | builds |
| Rsbuild 2.2.6 | `Conflict: ... same filename a.js` |
| 2.2.6 + `chunkIds: 'named'` | builds |
| 2.2.6 + `moduleIds: 'named'` | still conflicts |

So `chunkIds` alone is responsible. With `deterministic` the server output is unambiguous again:

```
dist/bundles/
  0.js  126.js  574.js    <- async chunks, numeric ids
  a.js  b.js              <- entries
```

## Fix

Pin `chunkIds` to `deterministic` for the server environment. Nothing caches these bundles, so compact ids buy nothing there; the web environment keeps the new default.

## Verification

- `packages/cli/builder` snapshot tests pass; the only snapshot change is `server.optimization.chunkIds`, and both `web` entries keep `compact-hashed`.
- Verified against the failing downstream build: applying the same server-scoped `chunkIds` makes it succeed.

Co-Authored-By: Riff


keepview:
Not needed after Rsbuild 2.2.7.

Rsbuild 2.2.7 bumps `@rspack/core` to 2.2.5, and that release stops compact-hashed chunk ids from landing on an entry name. Re-ran the exact repro from the description on `main` (no patch applied) with Rsbuild 2.2.7:

```
tests/integration/entries/fixtures/app-custom-entries, entry renamed to `b`

dist/bundles/
  1h.js 1.js 5.js 6.js 9.js ba.js b.js d.js r.js t.js u.js w.js x.js
  entry-2.js entry-2_page.js entry-2-server-loaders.js index.js
```

Builds clean; `b.js` is the entry and no async chunk claims that name. The same fixture failed deterministically on 2.2.6. A downstream two-entry SSG fixture that also failed on 2.2.6 now builds as well.

Closing. If a collision resurfaces, pinning `chunkIds` for the server environment is still the smallest fix and this diff can be revived.

PR #8881, opened 2026-09-17
feat(builder): accept full plugin options in tools.less / tools.sass and add tools.svgr

keepview:
## Summary

Modern.js assembled `@rsbuild/plugin-less`, `@rsbuild/plugin-sass` and `@rsbuild/plugin-svgr` with hard-coded options: `tools.less` / `tools.sass` were typed as the *loader* options only, and SVGR could not be configured at all. Plugin-level options such as `parallel` (worker-thread compilation, Rspack ≥ 1.4), `include`, `exclude` and `rewriteUrls` therefore had no entry point, even though the plugin versions in use already ship them.

This PR converges the entry point to `tools.<plugin>` taking the **full plugin options**, without breaking any existing config:

- `tools.less` / `tools.sass`: type widened to `PluginXxxOptions['xxxLoaderOptions'] | PluginXxxOptions`. The two shapes share no keys, so the runtime picks the layer by the keys present (`lessLoaderOptions` / `include` / `exclude` / `parallel` for less, `sassLoaderOptions` / `include` / `exclude` / `rewriteUrls` for sass). An object carrying a plugin-level key is passed through; anything else (object, function, array) is wrapped into `lessLoaderOptions` / `sassLoaderOptions` exactly as before, so the generated bundler config is unchanged.
- New `tools.svgr: ConfigChain<PluginSvgrOptions> | false`. User values are merged on top of the built-in defaults (`mixedImport: true`, `exportType` derived from `output.svgDefaultExport`); `svgrOptions` is merged one level deep so `svgrOptions.icon` keeps the derived `exportType`, while an explicit `svgrOptions.exportType` overrides it. `false` skips the plugin.
- `output.svgDefaultExport` and `output.disableSvgr` keep working and are marked `@deprecated`; `tools.svgr` wins when both are set.
- Migration hints for the legacy forms (and for objects that mix plugin-level keys with leftover loader options) are printed once per process, only by the `dev` / `start` commands, so `build` logs stay clean even with `NODE_ENV=development`.
- Docs (zh/en): `tools/less` and `tools/sass` rewritten for the new shape, new `tools/svgr` page, deprecation notices on the two `output` pages, SVG guide and plugin overview point at `tools.svgr`.

```ts
export default defineConfig({
  tools: {
    less: { parallel: true, lessLoaderOptions: { lessOptions: { javascriptEnabled: false } } },
    sass: { rewriteUrls: false },
    svgr: { parallel: true, svgrOptions: { exportType: 'default' } },
  },
});
```

## Verification

- `packages/cli/builder` unit tests: 67 passed / 0 failed on Node 22.22.0 (24 new cases: layer detection, legacy wrapping, mixed objects, svgr merge precedence and array chains, hint gating by command, and bundler-config assertions through `createBuilder().inspectConfig()`). The pre-existing `default` / `environment` snapshots are untouched.
- `modern inspect` on `tests/integration/ssr/fixtures/base` with a builder built from `main` vs. from this branch: the five emitted files (`modern.js.config`, `rsbuild.config.{client,server}`, `rspack.config.{client,server}`) are byte-identical both with the default config and with the legacy `tools.less: { lessOptions }` / `tools.sass: { sourceMap }` / `output.svgDefaultExport: 'component'` form. With the new form, all three less-loader uses and both svgr uses carry `parallel: true`, `resolve-url-loader` is dropped and `exportType: 'default'` is applied on client and server.
- Real CLI on the same fixture with legacy options: `NODE_ENV=development modern build` and `modern inspect` print no hint; `modern dev` prints the four hints.
- New integration fixture `tests/integration/tools-plugin-options` (string SSR app with `tools.less.parallel`, `tools.sass.rewriteUrls: false`, `tools.svgr` parallel + `exportType: 'default'`): runs a real `modern inspect` and asserts on both the client and the server rspack config (every less-loader / svgr use carries `parallel: true`, `lessLoaderOptions` reaches less-loader, resolve-url-loader is gone, `exportType` is `'default'`), then runs `modern build` and checks the Less / Sass CSS output. 7 passed / 0 failed locally on Node 22.22.0 (also reproduced independently by the reviewer). Note on CI scope: the framework integration workflows (`integration-test-Linux` / `integration-test-Windows`) only run on pushes to `main` / `renovate/main-rsbuild` or on manual dispatch, so they do not run for this PR; the numbers above are local runs. Reproduce from the repo root with:

  ```bash
  MODERN_SERVER_LOG_LEVEL=info pnpm --dir tests exec rstest run integration/tools-plugin-options
  ```

## Related Links

- RFC (internal): Modern.js / EdenX 插件配置面收敛为单一入口
- Rsbuild plugin docs: [plugin-less](https://rsbuild.rs/plugins/list/plugin-less), [plugin-sass](https://rsbuild.rs/plugins/list/plugin-sass), [plugin-svgr](https://rsbuild.rs/plugins/list/plugin-svgr)

## Checklist

Co-Authored-By: Riff


packages/document/docs/zh/guides/basic-features/static-assets/svg-assets.mdx
GiveMe-A-Name: 仍然通过 tools.bundlerChain 查找 RULE.SVG、ONE_OF.SVG 和 USE.SVGR。
建议把示例并改成 tools.svgr.svgrOptions.svgoConfig，英文文档也有同样的问题。

GiveMe-A-Name (approved):

packages/document/docs/zh/guides/basic-features/static-assets/svg-assets.mdx
keepview: done

